PR #6, opened 2026-06-02
New wrong_pin test and associated updates

tommelters:
## Summary
- New test `shared/flows/features/wrong_pin.yaml` — enters the wrong PIN ten
  times, closes each alert, asserts the wallet is wiped and lands on Signup1.
- PinViewController: lockout check moved inside the wrong-PIN branch.
- BitcoinManager.signMessage: guard self.ldkNode instead of force-unwrapping, to prevent crash during wallet reset.
- LoadWalletData: skip checkPendingPayout during a wipe/PIN reset.
- LoadWalletData: only reveal the euro + profit labels once the sats balance is known.
- Incidental: RestoreViewController styling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ maestro test shared/flows/features/buy_more.yaml
 maestro test shared/flows/features/receive.yaml
 maestro test shared/flows/features/swap.yaml
 maestro test shared/flows/features/remove_wallet.yaml
+maestro test shared/flows/features/wrong_pin.yaml
 
 # Forgot-PIN recovery test — needs the wallet's 12-word mnemonic so the
 # flow can type it on the RestoreVC screen. Pass it via --env:
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -18,10 +18,21 @@ extension BitcoinManager {
             if self.bdkWallet == nil {
                 Log.info("Will start blockchain and wallet.")
                 
+                // Bail out if the mnemonic has been cleared from the cache. This
+                // happens during a wallet wipe: performWalletReset deletes the
+                // mnemonic (deleteClientInfo) and clears bdkWallet (resetNodeState),
+                // so a didStartBDK that runs after teardown would force-unwrap a
+                // nil mnemonic and trap with a Swift _assertionFailure.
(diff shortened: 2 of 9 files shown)


RubenWaterman (commented):
## Review + follow-up fixes (commit 283746c)

Reviewed the PR and traced the lockout / wipe / label-gating changes through their call paths. The core change is a solid improvement — the lockout now only fires on a wrong entry and triggers exactly on the 10th, and I verified the counter resets on both correct entry and wipe (`deleteClientInfo` → `resetFailedPinAttempts`), so a freshly set-up wallet correctly starts at 0.

Pushed three follow-up fixes:

**1. Lockout-bypass edge — `PinViewController.swift`**
Changed the trigger from `== 9` to `>= 9`. The old code used `> 9` with the check *before* the increment, so the persisted `failedattempts` counter could legitimately be stored at 10. A user upgrading mid-abuse would have `10 != 9` and never lock out — the anti-theft wipe would silently stop working. `>=` keeps fresh-install behavior identical (caps at 9 internally) while catching any stuck-high counter.

**2. Wipe crash — `BDKManager.didStartBDK`**
A real `EXC_BREAKPOINT` crash on the wipe path (surfaced once the lockout actually fired). During a wipe, `performWalletReset` deletes the mnemonic and clears `bdkWallet`; a `didStartBDK` running after teardown then force-unwrapped a nil `getMnemonic()` and trapped. Replaced the force-unwrap with a graceful `guard`/bail, mirroring the existing `didSyncBdkWallet` teardown guard this PR already added.

**3. `wrong_pin.yaml` flow**
- Added `waitToSettleTimeoutMs: 0` on the digit taps — Maestro was waiting on each PIN button's background-flash animation across 40+ taps.
- Removed the `assertVisible`/`tapOn` of the lockout alert on the 10th attempt. That alert is shown and *synchronously* hidden by the wipe (`restoreWalletTapped` → `startWalletInBackground` → `hideAlert`), so it never reliably renders. The flow now waits for Signup1 directly. Confirmed passing end-to-end (10 wrong PINs → wipe → Signup1).

Non-blocking: `BitcoinManager` still has other unguarded `ldkNode!` force-unwraps (out of scope here); the one on the payout path was correctly fixed.

PR #7, opened 2026-06-02
Prevent wallet reset while Lightning channels are still open

tommelters:
- ResetApp: drop the channelWasClosedRecently() bypass (and its channelClosingInitiated/channelClosingTimestamp UserDefaults bookkeeping) that allowed a wallet reset to proceed within 2 minutes of initiating a channel close. A reset with open/closing channels could strand funds, so the close must now fully confirm before the reset is re-triggered.
- ResetApp: replace the two hardcoded English force-close alert strings with Language.getWord lookups (forceclose3 / forceclose4).
- Language: add forceclose3 / forceclose4 entries.


RubenWaterman (commented):
## Review + follow-up: gate the wallet reset on funds being safe (commit 244fc21)

Reviewed the original PR (removing the `channelWasClosedRecently` escape hatch) — it's a clean, correct safety improvement: I verified the removed symbols (`channelWasClosedRecently`, `channelClosingInitiated`/`channelClosingTimestamp`) have no remaining references, and the `forceclose3`/`forceclose4` string extraction is faithful.

On review we found the deeper risk and addressed it (full design discussion offline). Summary of what 244fc21 adds:

**The core problem.** Force-closing a channel and then wiping the wallet can *permanently* lose funds: the to_local output is CSV-locked (~1 day) and only swept later, and the sweep needs the LDK channel-monitor state that the wipe deletes — a BIP39 seed alone cannot reconstruct it.

**The fix (Option A):**
- `BitcoinManager.channelsFullyClosedAndSwept()` — one predicate for "safe to wipe" (no channels, no lightning balance, no pending closure sweeps; returns false if the node isn't running, so we never wipe blind).
- `ResetApp` — `performWalletReset` is now reachable **only** behind that gate. With an open channel we only ever **cooperatively** close (no auto force-close in the wipe path), then defer the reset behind a "still closing" / "retry later" alert; it completes on a later trigger once funds land. Manual force-close stays as an explicit user choice, gated identically.
- `PinViewController` — on the **3rd** wrong attempt, a warning explains the 10-attempt wipe can cost Lightning funds and offers a **Forgot PIN** button into the non-destructive mnemonic reset.
- `Restore3ViewController` — clears the failed-PIN counter on a successful mnemonic-verified reset (otherwise recovery left the user partway to a wipe).
- All new copy also warns never to delete the app (loss of funds).

**Test:** `shared/flows/features/pin_warning.yaml` — self-contained (runs `restore_wallet` first for a clean counter); verifies the warning appears at attempt 3 and recovery via its Forgot PIN button. Green locally.

### Known gaps / follow-ups (not blockers)
- **Open-channel path is not yet covered by an automated test.** The cooperative-close + deferred-wipe logic compiles and is exercised manually, but no Maestro flow drives a funded channel through it yet — tracked as a follow-up (needs a channel + regtest mining; `remove_wallet.yaml` is the template).
- **Coop-only trade-off:** if the bittr node is permanently offline, the automated wipe can't complete (funds aren't lost — manual force-close remains — but the locked-out user is stuck until the peer returns). Deliberate, per the funds-safety priority.
- This branch still shows `== 9` for the lockout; PR #6's `>= 9` arrives on merge into the base.

PR #8, opened 2026-06-05
Hardcoded Bittr articles

tommelters:
Replaces the network-backed article system with hardcoded content so signup and Academy article screens work fully offline.

- Add BittrArticles.json holding the 7 articles; getLocalArticle parses it
  (synchronous, returns Article?) replacing the API-based getArticle.
- setSignupArticle / getLocalArticle return Article? directly (no async /
  completion handlers); all 6 signup call sites updated.
- Bundle the 7 header images as asset-catalog imagesets keyed by slug;
  signup and ArticleViewController load UIImage(named: slug) instead of
  downloading + caching. getImage/CacheManager stay for Academy lessons.
- Point production Esplora endpoint at esplora.getbittr.com and route the
  onchain address check through EnvironmentConfig.esploraURL.


RubenWaterman (commented):
## Review + follow-up commit (9fbb7f3)

Reviewed the move from API-fetched articles to a hardcoded JSON + local asset images. It's clean and internally consistent — I verified the parts that could actually break:

- **JSON parses** (7 articles) and the raw-string literal `#"""…"""#` correctly preserves the JSON's own `\"` escaping.
- **Image keying lines up**: `parseArticles` sets `article.id` = the slug, so `ArticleViewController`'s `UIImage(named: article.id)` and `setSignupArticle`'s `UIImage(named: articleSlug)` both resolve to the imageset names.
- **All referenced slugs exist**: every signup page's `pageArticle1Slug` is in the hardcoded set with a bundled image — no page can land on an empty article.
- **No dangling references**: `getArticle`/`setArticleImage` fully removed, all 6 `setSignupArticle` call sites updated to the sync signature.
- **Escaped pseudo-tags render correctly**: `OneArticleTableViewCell` un-escapes `&lt;`→`<` then maps `<title>`/`<normal>`/etc. to styled spans.
- **esplora change is narrow**: `esploraURL` is consumed only by `ReceiveOnchain.checkHasBeenUsedByBittr` (no-op in prod; dev now hits `esplora.bittr.io` — worth a quick confirm that endpoint serves `/address/<addr>`).

### Follow-up changes pushed in 9fbb7f3
1. **Image weight: ~13 MB → ~1.7 MB.** The 7 header PNGs were full-res (1.4–2.3 MB each). Converted to **HEIC** (q80) — native to the 17.4 deployment target, and the illustrations have no alpha/text so it's artifact-free. (Not WebP — decodable on iOS 14+ but not idiomatic and weak asset-catalog tooling.)
2. **Removed dead data/code**: the unused S3 `headerimage` URLs and the now-readerless `headerimage` parse block (`Article.image` had no readers once images moved to the catalog). Left `getImage` — it turned out to be a shared `UIViewController` extension still used for Academy lesson images.
3. **Fixed baked-in content typos** (now that articles ship in the binary and can't be patched remotely): `July 3th` → `3rd` (×3), and the `wallet-recovery` article's title, which had been copy-pasted from the bitcoin-wallet article.

### Note on intent
Hardcoding removes the ability to update/add articles without an app release — presumably the deliberate trade for offline reliability, flagging since it's an architectural shift.

PR #9, opened 2026-06-05
Add bitcoin_value.yaml test for the ValueViewController

tommelters:
Cover ValueViewController end-to-end: unlock with the PIN, open the chart from Home's currency icon, wait for the price data to load, verify the latest conversion rate and percentage change, scrub the graph, and switch the span to 1 month / 1 year / 5 years.

Wire up the accessibility identifiers the flow needs: add home.currencyButton and a value.* block to test-ids.json (regenerated TestIDs.swift) and set them on the Home currency button, the Value labels/spinner/graph/span buttons, and the GraphView scrub value label. Document the flow in the READMEs and parity tracker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ maestro test shared/flows/features/receive.yaml
 maestro test shared/flows/features/swap.yaml
 maestro test shared/flows/features/remove_wallet.yaml
 maestro test shared/flows/features/wrong_pin.yaml
+maestro test shared/flows/features/bitcoin_value.yaml
 
 # Forgot-PIN recovery test — needs the wallet's 12-word mnemonic so the
 # flow can type it on the RestoreVC screen. Pass it via --env:
--- a/ios/bittr/Helpers/TestIDs.swift
+++ b/ios/bittr/Helpers/TestIDs.swift
@@ -22,6 +22,7 @@ enum TestID {
         static let balanceCardButton = "home.balanceCardButton"
         static let balanceLabel = "home.balanceLabel"
         static let buyButton = "home.buyButton"
+        static let currencyButton = "home.currencyButton"
         static let headerLabel = "home.headerLabel"
         static let headerSpinner = "home.headerSpinner"
         static let receiveButton = "home.receiveButton"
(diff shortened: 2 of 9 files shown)


RubenWaterman (commented):
## Review + follow-up commit (5bf8302)

A test-only PR — adds accessibility identifiers, registers the matching test IDs, and adds a Maestro flow for the price-chart screen. Reviewed and it's clean:

- **Compiles**: every outlet getting an `accessibilityIdentifier` exists (`currencyButton` on HomeVC; all 8 on ValueVC).
- **No production logic changed** — only identifier assignments.
- **Test IDs consistent**: `shared/test-ids/test-ids.json` (`value` + `home.currencyButton`) exactly mirrors the `TestID.Value` Swift enum.
- **Docs correct**: `parity.md` and both READMEs updated.
- `graphValueLabel` is assigned to a single scrub card per `showGraphValue` call (computed from one `selectedIndex`), not a loop — no id collision.

### Follow-up: fixed a flaky data-load wait (5bf8302)
The flow failed on the `value.profitLabel` wait — the screen was still spinning at 30s. Not an accessibility issue: `getCurrentValue` makes **two sequential calls** to the live getbittr API (historical chart ~5s + current price ~9s, both variable) and then parses ~120KB, building a fresh `ISO8601DateFormatter` per data point across the week/month/year/5y arrays. On a debug build + simulator that routinely exceeds 30s. Raised the `extendedWaitUntil` timeout to **90s** (it returns as soon as the element appears, so the higher ceiling only costs time on a genuine failure). Confirmed both endpoints return 200.

### Notes (non-blocking)
- **Live-API dependency**: even at 90s this fails if getbittr.com is down or latency spikes — inherent CI-flakiness risk since the screen isn't mocked. Worth keeping in mind for the Android parity run.
- **Span-switch assertions are shallow**: after m/y/5y, the flow only re-asserts `value.graphView` (always present), so it documents via screenshots rather than verifying the span reloaded. Fine, just noting.
- **Pre-existing app smell (out of scope)**: the per-data-point `ISO8601DateFormatter()` in `getCurrentValue` is needlessly expensive — a single shared formatter would cut most of the parse time and shrink the load wait. Cheap future win, unrelated to this test PR.

RubenWaterman:
Folded the date-parsing cleanup I'd flagged into this PR (ec7ac50): `getCurrentValue` now reuses a single `ISO8601DateFormatter` and pre-computed span cutoffs, parsing each timestamp once instead of building two formatters per data point across all four spans. Behavior-preserving — a malformed timestamp now skips the point (guard) instead of trapping on a force-unwrap.

Note: total load time is still network-dominated (~13s of sequential API calls), so this trims the parse spike rather than collapsing it — the 90s wait stays as the safe ceiling for API latency variance.

PR #10, opened 2026-06-05
Add bitcoin_map.yaml test for MapVC

tommelters:
Cover MapViewController + OnePlaceViewController end-to-end: unlock with the PIN, open the map from Home's map icon, wait for the btcmap places to sync, verify the places table is populated, open the topmost place and close it, move the map, and recentre on the user's location.

Wire up the accessibility identifiers the flow needs: add home.mapButton and a map.* block (incl. nested onePlace) to test-ids.json (regenerated TestIDs.swift) and set them on the Home map button, the map view / spinner / places table / user-location button, the place cells, and the OnePlace name and close button. Document the flow in the READMEs and parity tracker.

Location is granted up front via launchApp.permissions (location: inuse) so no system permission dialog interrupts the flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ maestro test shared/flows/features/swap.yaml
 maestro test shared/flows/features/remove_wallet.yaml
 maestro test shared/flows/features/wrong_pin.yaml
 maestro test shared/flows/features/bitcoin_value.yaml
+maestro test shared/flows/features/bitcoin_map.yaml
 
 # Forgot-PIN recovery test — needs the wallet's 12-word mnemonic so the
 # flow can type it on the RestoreVC screen. Pass it via --env:
--- a/ios/bittr/Helpers/TestIDs.swift
+++ b/ios/bittr/Helpers/TestIDs.swift
@@ -25,8 +25,21 @@ enum TestID {
         static let currencyButton = "home.currencyButton"
         static let headerLabel = "home.headerLabel"
         static let headerSpinner = "home.headerSpinner"
+        static let mapButton = "home.mapButton"
         static let receiveButton = "home.receiveButton"
         static let sendButton = "home.sendButton"
+    }
+    enum Map {
+        static let mapSpinner = "map.mapSpinner"
+        static let mapView = "map.mapView"
+        enum OnePlace {
+            static let closeButton = "map.onePlace.closeButton"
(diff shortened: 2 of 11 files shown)


RubenWaterman (commented):
## Review + follow-up commit (901ad87)

Same shape as the bitcoin_value PR — test IDs, accessibility-identifier assignments, and a Maestro flow for the map screen. Reviewed; the code side is clean:

- **Compiles** — every assigned outlet exists (`mapButton` on HomeVC; `mapView`/`mapSpinner`/`placesTableView`/`userLocationButton` on MapVC; `placeNameLabel`/`closeButton` on OnePlaceVC; the cell's `placeName`/`cellButton` already used in unchanged code).
- **No production logic changed** by the test scaffolding — only identifier assignments.
- **Test IDs consistent** — `test-ids.json` `map` (incl. nested `onePlace`) mirrors `TestID.Map`; `home.mapButton` in both.
- The per-cell shared ids (`placeName`/`placeCellButton`) are handled correctly in the flow via `index: 0` and `assertVisible`.

### Follow-up: fixed a real hang the test surfaced (901ad87)
Running the flow hung the app on the default map view. Root cause is in the screen, not the test: `showDefaultSwitzerlandRegion` uses a ~422 km radius, so `nearbyBTCPlaces` matched thousands of places and `showPlacesOnMap` added an `MKAnnotation` for each plus reloaded a table that long. Capped `currentPlaces` to the **nearest 200** to the map centre; the existing 0.6s region-change debounce recomputes it as the user pans/zooms, so the visible set is always the 200 nearest to where they're looking. Bounded at any zoom level, no interaction state to reason about.

### Notes (non-blocking)
- **Live `btcmap.org` dependency** — the 120s `mapSpinner` wait is appropriately generous for a first-ever full sync, but it's an external service, so it's a CI-flakiness vector (same caveat as the price API in the value PR).
- **Location determinism** — with the cap in place the Switzerland fallback no longer hangs, so the flow no longer depends on the Simulator's location state to populate the table. (Granting `location: inuse` would still recenter to a tight 1500 m window if a location is simulated, but the table stays bounded either way.)
- `assertNotVisible: map.onePlace.nameLabel` after close assumes OnePlaceVC is torn down rather than slid offscreen — verified visually in the run.

PR #12, opened 2026-06-09
New receive_onchain.yaml test and associated updates

tommelters:
- Fixes the onchain receive-address race (background verifier is the single cache writer; read-only display path; alertTapped gate + onchainAddressesReady notification; bounds guard in getNextUnusedAddress).
- Adds the receive_onchain Maestro flow + show_onchain_address helper, new receive/send test IDs (regenerated TestIDs.swift), and README/parity doc updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ maestro test shared/flows/features/buy_incoming.yaml
 # Subsequent feature tests reuse that channel:
 maestro test shared/flows/features/buy_more.yaml
 maestro test shared/flows/features/receive.yaml
+maestro test shared/flows/features/receive_onchain.yaml
 maestro test shared/flows/features/swap.yaml
 maestro test shared/flows/features/remove_wallet.yaml
 maestro test shared/flows/features/wrong_pin.yaml
--- a/ios/bittr/Base.lproj/Main.storyboard
+++ b/ios/bittr/Base.lproj/Main.storyboard
@@ -4221,7 +4221,7 @@
                                                                     <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="O5E-tG-CSf" userLabel="Address View">
                                                                         <rect key="frame" x="0.0" y="0.0" width="323" height="45"/>
                                                                         <subviews>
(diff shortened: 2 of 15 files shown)


RubenWaterman (commented):
## Review + follow-up commit (3611254)

This PR adds the receive_onchain flow plus real logic changes (address race fix, `getNextUnusedAddress` rewrite, funding-tx dedup). Reviewed in depth.

### The good
- **`getNextUnusedAddress` rewrite fixes a latent crash** — the old `while !resultHasBeenFound` had no lower bound, so with no cached/used match `checkIndex` went negative and `self[-1]` would trap. The new `while checkIndex >= 0` + explicit `return nil` is correct, and every caller handles `nil` (swap paths fall back to `getAddress(atIndex: 0)`, ReceiveVC via `??` chains).
- Test-ID scaffolding compiles (outlets verified) and is consistent; the `bcrt1` regtest address pattern is a correct addition.

### Finding #1 [Medium] — onchain Receive could hang on a perpetual spinner — fixed in 3611254
The new display gate bails and waits for `onchainAddressesVerified`, but that flag is only set on one completion path. Two cases left it `false` forever with no recovery: `revealOnchainAddresses()` early-returns when `getNewOnchainAddress()` is nil (before the flag is set), and `manageOnchainAddresses()` may never run if startup didn't reach it. Fixed at the source (the early-return now marks verification done + notifies) plus an 8s timeout fallback in `alertTapped` that shows the best-available cached address. The timeout no-ops on the normal fast path and won't fire after the user switches receive type.

### Finding #2 — retracted (not a bug)
My earlier "funding-tx double-add" note was an over-call (I'd flagged it as "worth confirming"). On re-reading: the funding-tx line is in the `else` branch, where `visibleTransactions` is **not** pre-added (that loop is only in the `sendAll` branch); the channel-funding tx isn't in the lightning `allTransactions` list either, so there's no double-add. `sendableTxIDs` only feeds a bittr API *query* where a duplicate would be harmless regardless. The `!(sent && visible)` logic is an intentional "re-send the funding tx until it's confirmed visible." No change needed.

### Notes (non-blocking)
- `onchainAddressesVerified` is never reset for the session — fine for initial-display gating.
- Dev `lightningNodeId` changed (regtest only; prod untouched) — config; confirm it's the intended regtest node.
- Cosmetic: the testnet regex comment gained a stray trailing comma (`Testnet,`).

PR #13, opened 2026-06-11
New receive_invoice.yaml test

tommelters:
- Mirrors receive_onchain.yaml for the lightning invoice type: open Receive, switch to a lightning invoice (More → Create invoice) via the new helpers/show_invoice.yaml subflow, read its info and copy it, paste into Send asserting it lands as lightning with no amount, then attach a 2000 sat amount and verify the invoice + amount carry across.
- Wire up the missing test ID for the invoice: the invoice string renders into lowerAddressLabel, which had no accessibilityIdentifier. Add receive.invoiceLabel to test-ids.json, regenerate TestIDs.swift, and set it on the label.
- Also wait for the Home sync spinner before interacting with the balance card in swap.yaml. Document the new flow/helper in the main README, flows README, and parity tracker.


RubenWaterman (commented):
## Review — clean, mergeable

Essentially a test-only PR.

- **Production change is a single line** — `lowerAddressLabel.accessibilityIdentifier = TestID.Receive.invoiceLabel`. The outlet exists (`ReceiveViewController:37`), so it compiles with zero behavioral impact.
- **Test IDs consistent** — `invoiceLabel` added to both `test-ids.json` and the `TestID.Receive` enum.
- **New flows are well-built** — `show_invoice.yaml` / `receive_invoice.yaml` reference only valid IDs, mirror the `receive_onchain` structure, gate on the QR spinner, and the `text: ".*lnbcrt.*"` assertion works against a normal `UILabel`.
- **`swap.yaml` hardening** — the two added `extendedWaitUntil notVisible home.headerSpinner` waits are purely additive robustness (wait for sync before tapping the balance card / checking `swapPending`). No regression.

### Non-blocking notes
- `show_invoice.yaml` hard-codes `alert.button.3` = "Create invoice" (type-picker order `[Cancel, Address, Bitcoin QR, Create invoice, Show LNURL]`) — repo convention, but brittle if the picker is reordered.
- **Requires an active lightning channel** — without one, Receive opens in plain onchain Address mode (no "More" type-picker), so the flow fails at `receive.moreButton`. Run `buy_incoming.yaml` first. (Confirmed via a run; it's a prerequisite, not a flow bug.)
- Regtest-scoped by design (`lnbcrt` prefix).

No fixes needed.

PR #14, opened 2026-06-11
New send_onchain.yaml test and associated improvements

tommelters:
Add features/send_onchain.yaml: an end-to-end onchain send — open Send, switch to Regular (waiting out the BDK sync spinner), enter an address and a 5 EUR amount, confirm the address/euro amount on the Confirm screen with the fast fee, send, mine 6 blocks, and open the new transaction from the history table. The flow remembers the previous top transaction and polls until the new one replaces it, so it never taps the wrong row.

Supporting changes:
- Wire up the missing test IDs: send.regularButton/currencyButton/currencyLabel/ bdkSpinner, the send.confirm.* fields, and row-indexed history.transactionButton/ transactionAmount so the topmost cell is addressable as ...0.
- Fetch the latest block height during the Home pull-to-resync so transaction confirmation counts reflect newly mined blocks (startWallet didn't fetch it).
- Show the onchain Confirm fiat amount with two decimals.
- Show the confirm alert amount in satoshis instead of a tiny BTC value.
- Replace the Send currency action sheet with the shared custom alert used by ReceiveViewController's More picker.


RubenWaterman (commented):
## Review + follow-up commit (8d43cc1)

Beyond the send_onchain flow, this PR has real Send/Confirm/Home changes. Reviewed in depth — clean, no correctness bugs.

### Verified sound
- **`sendconfirmation` amount** — the `<amount>` value changed from `.inBTC()` to raw `confirmSatoshis`, *and* the template was updated in the same PR ("`<amount>` BTC" → "`<amount>` satoshis"). Consistent — the confirmation correctly reads "send N satoshis". (Top thing to verify; it's fine.)
- **`formattedFiatAmount` doesn't crash** — `fixDecimals` maps `.`/`,` to the locale separator and the interpolated `Double` always has a `.`, so `split[1]` is never empty; `split(separator:)` resolves to the literal `RegexComponent` overload.
- **Compiles** — all new outlets exist (Send: `regularButton`/`btcButton`/`btcLabel`/`bdkSpinner`; Confirm: `addressLabel`/`amountFiatLabel`/`buttonFast`/`confirmButton`; History cell: `transactionButton`/`satsLabel`), `didGetLatestBlockHeight()` exists, and test-IDs ↔ Swift enums match.
- Row-indexed history IDs are set per-render (reused cells stay correct); the `btcButtonTapped` switch from a native action sheet to the shared custom alert wires buttons↔actions correctly (`selectcurrency` keys added).
- `ReloadWallet` refreshing the chain tip before `startWallet()` on the reset path is reasonable (confirmation counts) at the cost of one extra call.

### Follow-up fixed in 8d43cc1 — fiat decimal separator not localized
Both `formattedFiatAmount` (send amount) and `convertFees` (mining-fee fiat) built a dot-decimal string and returned it **raw**, so the confirm screen showed "4.99 €" / "0.50 €" regardless of locale — inconsistent with the rest of the app, which renders the device separator (e.g. "4,99" / "0,50" for comma locales, i.e. most EU/CH users). Now both run the final string through `fixDecimals()`. Also relaxed the `send_onchain.yaml` fiat-amount regex from `[45]\.[0-9][0-9] €` to `[45][.,][0-9][0-9] €` so it passes on both locales (the original failure was the test demanding a "." while the fixed label correctly showed ",").

### Note
The fiat formatting is still hand-rolled (`CGFloat(Int(x*100))/100` + pad); a currency `NumberFormatter` would be cleaner (rounding instead of truncation, thousands grouping) — non-blocking, out of scope here.

PR #15, opened 2026-06-11
New send_onchain_all.yaml test and associated improvements

tommelters:
Add features/send_onchain_all.yaml: sends the maximum onchain balance via the "Send all" label and exercises the tight-fee path — the fast fee exceeds the balance (Update amount) and a failed broadcast retries with the slowest fee (past the low-fee warning) — before mining and opening the new transaction. Optional alerts are branched on their (plain-text) titles.

Supporting changes:
- Wire up the missing test IDs: send.availableButton, send.confirm.amountLabel, send.confirm.feeSlowButton.
- Sync the BDK wallet during the Home pull-to-resync, so the onchain wallet used for fee/size calculation and sends picks up newly received/confirmed funds instead of going stale until the next BackgroundSync tick.
- Unwrap LDKNode errors in the onchain send failure alert via the shared handleNodeError handler, so an insufficient-funds failure shows the plain message rather than the raw NodeError enum description.


RubenWaterman (commented):
## Review — clean, mergeable

Production "Send fixes" verified sound:

- **`bdkWalletIsScanning` guard** (`SendViewController`, both sites) — treats the send-all amount as unavailable *while* a BDK scan is in flight, fixing the stale/mid-scan max-sendable amount. The flag (`BitcoinManager.swift:24`) is reset to `false` in every `didSyncBdkWallet` completion/error path, so it can't get stuck.
- **`SendOnchain` error unwrap** — `handleNodeError(nodeError).detail` surfaces the human-readable LDK message instead of the raw `NodeError` enum. Both symbols exist; compiles; UX win.
- **`ReloadWallet`** — refreshes `setSendAllLabel()` + `calculateSendableAmount()` after pull-to-refresh; all optionals/methods resolve.
- **Test IDs** (`amountLabel`, `feeSlowButton`, `availableButton`) — outlets exist, `test-ids.json` ↔ Swift enums consistent. The new flow asserts the BTC `amountLabel` (self-compared), so it has no hard-coded fiat decimal-separator regex.

### Non-blocking notes
- **Possible redundant BDK full-scan on pull-to-refresh** — `ReloadWallet` calls `startWallet()` then `didSyncBdkWallet { … }`; if `startWallet` already scans BDK, the second scan exists mainly to get the completion callback for `setSendAllLabel()`. Calling `setSendAllLabel()` after `startWallet` (no re-scan) would be cheaper if so. Worth confirming.
- The `didSyncBdkWallet` closure captures `self` strongly, but it's a one-shot callback — transient retain, not a leak.
- Heads-up: this branch predates the PR #14 locale fix, so `send_onchain.yaml`'s fiat regex updates when #14 merges down; `send_onchain_all.yaml` is unaffected.

PR #16, opened 2026-06-12
New send_lightning.yaml test

tommelters:
End-to-end lightning send flow covering a normal (amount-bearing) invoice, a zero-amount invoice (amount entered in BTC), and a Lightning Address (LNURL-pay). Adds send.availableLabel/nextButton/lnurlSpinner test IDs and a clipboard bridge (clipboard_server.js + set_clipboard.js) so the in-app Paste button can be exercised. Three targets are manual input via LN_INVOICE / LN_ZERO_INVOICE / LN_ADDRESS env vars.

Also changes LNURL-pay behaviour: when minSendable != maxSendable the payable range is now shown via a 'Pay request' alert (title payrequest, message payrequest1) instead of being written into the send-all label.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,12 @@ maestro test shared/flows/features/send_onchain.yaml
 maestro test shared/flows/features/send_onchain_all.yaml
 maestro test shared/flows/features/swap.yaml
 
-# Lightning send (normal + zero-amount invoice + Lightning Address). Manual
-# input: the three targets come from a separate live wallet you run yourself.
+# Lightning send (normal + zero-amount invoice + Lightning Address). All three
+# targets are server-side now (invoices via the e2e endpoint / request_invoice.js,
+# address = fixed e2e e2ebittr@staging.getbittr.com), so no --env is needed.
 # Needs "node shared/flows/scripts/clipboard_server.js" running (Terminal A, alongside push_server)
 # so the in-app Paste button has something to paste:
-maestro test --env LN_INVOICE="lnbcrt…" --env LN_ZERO_INVOICE="lnbcrt…" --env LN_ADDRESS="name@domain" shared/flows/features/send_lightning.yaml
+maestro test shared/flows/features/send_lightning.yaml
 
 maestro test shared/flows/features/remove_wallet.yaml
 maestro test shared/flows/features/wrong_pin.yaml
--- a/README.md
+++ b/README.md
@@ -48,6 +48,13 @@ The push-notification helper bridges Maestro to the simulator's APNS push (the `
 node shared/flows/scripts/push_server.js
 ```
 
+The lightning send flow additionally needs the clipboard helper, which bridges Maestro to `xcrun simctl pbcopy` so the in-app Paste button has something to paste (see the `send_lightning.yaml` run line below). Start it the same way:
+
+```sh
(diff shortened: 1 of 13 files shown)


RubenWaterman (commented):
## Review + follow-up commit (8a0507b)

### Original PR — clean, no correctness bugs
- **`SendLNURL` pay-request range handling** is sound: when min≠max sendable it stores the pending callback/desc/min/max, clears the amount field, and shows the range via the new `payrequest` alert; the entered amount feeds `handleLNURLAmountCompletion`.
- **Units correct** — `enteredAmount` is converted to **msat** (`* 1000`) in all three currency branches and compared against the msat `minAmount`/`maxAmount`. No sats/msat off-by-1000.
- `payrequest`/`payrequest1` Language keys exist with matching placeholders; `nextButton`/`lnurlSpinner`/`availableAmount` outlets exist; test-ids ↔ Swift enums consistent. Compiles.
- Minor (non-blocking): the rewrite drops the persistent range label + `becomeFirstResponder()` in favor of the transient alert; the alert says "satoshis" while the field can be fiat/BTC — but the validation converts correctly. Cosmetic.

### Follow-up: targets now generated server-side (8a0507b)
The flow originally needed three hand-supplied `--env` values from a live counterpart wallet — tedious and flaky (bolt11s expire between runs). Now all three are server-side, so `send_lightning.yaml` runs with **no `--env`** (still needs `clipboard_server.js` for the in-app Paste button):
- **Invoices** — `request_invoice.js` fetches a fresh bolt11 from `POST /api/e2e/invoice {amount_in_sats}` (1000-sat + zero-amount). I curl-verified the endpoint: `200`, bolt11 returned in `payment_request` (the script also accepts `invoice`/`bolt11`).
- **Lightning Address** — fixed e2e `e2ebittr@staging.getbittr.com`.

Biggest win is eliminating invoice-expiry flakiness and making the flow CI-runnable unattended. Still depends on an active channel with outbound capacity for the pay step (documented).

RubenWaterman (commented):
## Addendum — also includes a Receive lightning-address fix (a9c512f)

Folding in a related bugfix to the Receive screen (lightning address showed "Unavailable" despite the backend returning `lightning_address_username`). Two distinct issues:

1. **Read side (likely the main culprit).** `userLNURL()` only checked `ibanEntities.first`, but `CacheManager.parseDevice` (which builds `bittrWallet.ibanEntities` on every load, `CoreViewController:134`) **sorts** the entities — so with more than one deposit code, the lightning-address iban may not be first and the address was missed even when present. Now it scans **all** entities for the first non-empty address.

2. **QR for an unavailable address.** `setLabels` built `lnurl.toBigQRCode()!` even when `lnurl` was the literal word "Unavailable" — encoding that word into a QR. Now `qrCode` is optional: no QR is generated when the address is unavailable, and the QR view is hidden (`alpha = 0`) instead of showing a meaningless code.

Note: there's also a deeper population caveat — the live API's `lightning_address_username` only reaches `bittrWallet`/cache via `BuyViewController.parseNewData`'s conditional sync; nothing else populates it. If the address is genuinely absent from the wallet (never cached), the read-side fix can't surface it. If it still shows "Unavailable" after a rebuild, that sync path is the next place to look.

PR #17, opened 2026-06-15
New settings.yaml test and SettingsVC improvements

tommelters:
- Moved "Currency" and "Remove wallet" into Device Details VC.
- Removed "Wallet and balance" from Settings VC.
- UI update: Settings VC is now a popup off of the CoreVC.
- New settings.yaml test which tests all SettingsVC and DeviceVC features.
- Custom alerts for currency and language selection in DeviceVC.


RubenWaterman (commented):
## Review — settings refactor is solid; + a swipe-to-dismiss addition (741e317)

### Original PR — no correctness bugs, mergeable
Verified the parts that could have broken in the Settings popup/navigation refactor:
- All new outlets (`settingsBackground`/`settingsContainer`/`settingsContainerTop`/`centerView`) are wired in the storyboard — no nil-outlet crash.
- Clean removals: the storyboard no longer wires the deleted `currencyLabel` IBOutlet, and the removed `changeCurrency` observer isn't referenced — no crash/compile break.
- `selectlanguage`/`selectcurrency` (+message) Language keys exist; the DeviceVC custom-alert pickers are consistent with the established pattern.
- Device-details "Remove wallet" → `restoreWalletTapped()` + `dismiss()` works: `restoreWalletTapped` uses the *custom* (subview-based) alert, so dismissing DeviceVC reveals it rather than colliding with a modal presentation; `performWalletReset` also calls `hideSettings()`.

Low-severity notes (none blocking):
1. `addHeader` is re-invoked from `changeColors` (a `changecolors` observer), with no dedup — but each header pins to the identical frame, so it overlaps pixel-for-pixel and isn't visible; the re-call is also how the header picks up the new theme's icons. Not a visible defect.
2. The settings child VC is removed from its superview but never `removeFromParent()`'d — this matches the existing `hideSignup` pattern (house style), not a regression.
3. `showSettings` reads `centerView.bounds.height` for the slide offset; fine given the preceding `layoutIfNeeded()`.

### Addition — swipe-down-to-dismiss (741e317)
There was no existing swipe-to-close anywhere (only `GraphView`'s chart-scrub pan), so this is new. A `UIPanGestureRecognizer` on `settingsContainer` tracks the finger via `settingsContainerTop` (clamped to downward-only from the open position) and fades the dim as it's pulled; on release it dismisses past ⅓ height or a fast flick, else springs back. It recognises simultaneously with the settings table's scroll pan so a drag started on a row still dismisses. Reuses the existing `hideSettings()` so dismissal continues smoothly from the drag.

Note: a "process launched but exited" with no crash report (5-min Xcode timeout) we hit while testing was a debugger/simulator-runtime issue (device iOS 18.5 vs SDK 26.2 under Xcode 26.3), not this change.

PR #18, opened 2026-06-16
Fixes to ProfitVC and testing

tommelters:
- Recalculate profits correctly upon receiving new payments through ldkEventReceived.
- Change language for Remove Wallet flow. Call it "remove" instead of "restore", since we're removing this current wallet - not restoring another wallet.
- Include ProfitVC testing in buy_incoming.yaml and buy_more.yaml tests.
- Design improvements for ProfitVC


RubenWaterman (commented):
## Review — clean, mergeable; fixes a real profit bug

- **"Profit stuck at 0% after the first buy" is genuinely fixed.** The old `calculateProfit` used a count-based terminator (`handledTransactions == bittrTransactions.count`) to decide when to render. Funding transactions are Bittr purchases that live in `visibleTransactions` but never in the `bittrTransactions` dict (they aren't LDK payments), so the terminator was never reached and the label stuck at 0%. The rewrite iterates `visibleTransactions where isBittr` and calls `showProfitLabel` once afterward — same math, no fragile terminator, empty case still shows 0%.
- **"Recalculate profit on new payments" is wired correctly** — the handler adds the transaction to `visibleTransactions` (line 319) before calling `calculateProfit()` (line 353), so the recalc includes it; only `isBittr` purchases count.
- **Compiles / no crashes:** the removed `divestedView` IBOutlet is no longer wired in the storyboard (no KVC crash); `yellowCard` + all Profit/Home outlets exist; the reworded alerts' new `remove`/`removewallet` Language keys are present; test-ids ↔ Swift enums match.
- ResetApp restore→remove rewording and the ProfitVC layout redesign are cosmetic and consistent.

Non-blocking note: `bittrTransactions.updateValue` moved to the top of the loop so funding txs now also enter that dict — harmless since `calculateProfit` no longer reads its `.count`, and no other consumer assumes funding is excluded.

No fixes needed.

PR #19, opened 2026-06-19
Fix keyboard flicker when moving between seed-phrase fields

RubenWaterman:
## What

Stops the keyboard from appearing to dismiss and re-open every time you move between word-entry fields on:

- **Wallet restore** and **Forgot PIN** recovery — `RestoreViewController` (shared by both flows)
- **New-wallet backup confirmation** (the 3 words you re-enter) — `Signup4ViewController`

## Why

Both screens observed **both** `keyboardWillShow` and `keyboardWillHide`. When focus moves between two text fields, iOS posts a hide *immediately followed by* a show even though the keyboard never actually leaves. The handlers reacted to both: `willHide` snapped the content view's bottom constraint back to `0`, then `willShow` shoved it back to `-keyboardHeight`. That un-animated down-then-up bounce is the "keyboard kind of re-appears" flicker.

Two aggravating factors: the constraint change wasn't animated (hard jump), and the observers were re-added on every `viewWillAppear` without removal, so the handlers multiplied on revisits.

## How

- Replace the `willShow`/`willHide` pair with a single **`keyboardWillChangeFrame`** observer.
- Compute the keyboard's actual overlap with the view and **bail out when it's unchanged** — which is exactly the field-to-field case, so the layout no longer churns.
- Animate the adjustment with the keyboard's own duration.
- Remove the observer before re-adding it so repeated `viewWillAppear` calls don't stack duplicates.

For these full-screen view controllers the computed overlap equals the previous `keyboardHeight`, so the resting content position is unchanged — only the flicker is gone.

## Testing

⚠️ Not compiled — built in the sandbox isn't possible here, so **please build in Xcode** and verify:
- Restore wallet: tab/return and tap between all 12 fields — keyboard stays put, no bounce.
- Forgot PIN (reset PIN flow): same fields, same check.
- New wallet → backup confirmation: move between the 3 fields.
- Keyboard still shows/hides correctly on entering/leaving the screen, and content isn't hidden behind the keyboard.

### DIFF
--- a/ios/bittr/Signup/Create Wallet/Signup4ViewController.swift
+++ b/ios/bittr/Signup/Create Wallet/Signup4ViewController.swift
@@ -103,8 +103,14 @@ class Signup4ViewController: UIViewController, UITextFieldDelegate {
     override func viewWillAppear(_ animated: Bool) {
         
         // Set notification observers.
-        NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillDisappear), name: UIResponder.keyboardWillHideNotification, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillAppear), name: UIResponder.keyboardWillShowNotification, object: nil)
+        // Use a single keyboardWillChangeFrame observer instead of the
+        // willShow/willHide pair: when focus moves between the word fields iOS
+        // posts a hide immediately followed by a show, and reacting to both
+        // bounced the content down then back up — the keyboard appeared to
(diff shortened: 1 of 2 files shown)


RubenWaterman:
Closing — not a real bug. The keyboard 'hides and re-appears' between seed-phrase fields only under Maestro (the XCUITest/idb automation re-focuses each field programmatically); it's smooth in normal use, so real users are unaffected. No app change warranted on the wallet-restore path. If the flicker turns out to cause test flakiness, the fix belongs in the Maestro flow, not the app.

PR #20, opened 2026-06-19
Updates to wrong_pin.yaml test and associated changes

tommelters:
With our recent changes to wallet removal, for a locked out user (after 10 incorrect pin attempts), it may take significant time before their wallet can be successfully removed from the device. The following changes are in support of that.

- I've fixed an issue where, after 10 incorrect pin attempts, a successful 11th attempt would still reinstate the user's access to their wallet. So users are now definitively locked out.
- Upon app launch, the app checks immediately whether the user has been locked out.
- For locked out users, the PinViewController fades out, alerts the user, and retries removing the wallet. 
- As long as the wallet cannot be removed, the only thing the locked out user can do, is tap "Try again" to retry removing the wallet.
- Include these changes in the wrong_pin.yaml test.
- Reset pinCollectionView after each incorrect pin attempt to empty the visual text field.
- When a channel has been closed as a result of incorrect pin, do not show the QuestionViewController with the ldkEventReceived notification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,11 @@ maestro test shared/flows/features/swap.yaml
 maestro test shared/flows/features/send_lightning.yaml
 
 maestro test shared/flows/features/remove_wallet.yaml
+# PIN lockout, no open channel (immediate wipe):
 maestro test shared/flows/features/wrong_pin.yaml
+# PIN lockout with an open channel (close + Try again retry loop) — set up a
+# channel first (e.g. run buy_incoming.yaml):
+maestro test shared/flows/features/wrong_pin_with_channel.yaml
 maestro test shared/flows/features/bitcoin_value.yaml
 maestro test shared/flows/features/bitcoin_map.yaml
 maestro test shared/flows/features/academy.yaml
--- a/ios/bittr/Core/CoreViewController.swift
+++ b/ios/bittr/Core/CoreViewController.swift
@@ -196,6 +196,20 @@ class CoreViewController: UIViewController {
         if self.walletIsAvailable {
             self.signupContainerView.alpha = 0
             self.pinContainerView.alpha = 1
+
+            // Check whether the user has been locked out of their wallet.
+            if CacheManager.getFailedPinAttempts() >= 10 {
+                Log.info("Wallet is locked out after 10 failed PIN attempts — resuming removal on launch.")
+                self.removingWalletForIncorrectPin = true
+                self.fullViewCover.alpha = 0.8
+                self.genericSpinner.startAnimating()
+                self.showAlert(presentingController: self, title: Language.getWord(withID: "restorewallet"), message: Language.getWord(withID: "pinlock"), buttons: [Language.getWord(withID: "okay")], actions: nil)
+                
+                // Start wallet in background.
+                Task {
+                    await self.startWallet()
+                }
+            }
         } else {
             self.signupContainerView.alpha = 1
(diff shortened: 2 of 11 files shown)


RubenWaterman (commented):
## Review — PIN-lockout / wallet-removal hardening

Reviewed the full diff; this is a solid hardening of the lockout flow. The core fixes are right: definitive lockout (a correct PIN after 10 failures still removes, rather than restoring access), the counter now reaches 10 (increment-before-check), launch-time resume, the `stillclosing` "Try again" retry, clearing the PIN field, and suppressing the channel-closed `QuestionViewController` during the wipe.

Four findings; all now resolved or accounted for (the two code fixes are pushed to this branch):

**1. (Medium) `closeretrylater` dead-end — fixed (`8b70857`).**
In `closeChannelConfirmed()`, when the automated lockout wipe can't cooperatively close the channel (peer offline, or `closeChannel()` throws), the spinner stopped but the full-screen cover was left up with only an `[okay]` button (`actions: nil`). Tapping it stranded the locked-out user on a static cover — can't enter a PIN (cover blocks touches), no "Try again" — recoverable only by relaunching. Both branches now show a **"Try again"** button wired to `startWalletInBackground`, matching the adjacent `stillclosing` handling.

**2. (Low) Launch re-entry guard — fixed (`8b70857`).**
`showPinOrSignup()` raised the lockout alert + a `startWallet()` task with no guard. Added `!removingWalletForIncorrectPin` so a repeat call can't stack a second alert or spawn a concurrent `startWallet()`.

**3. (Low) `startWallet()` vs `startWalletInBackground()` — no action.**
On closer look the divergence is intentional: the launch path keeps the just-shown lockout alert visible, which `startWalletInBackground`'s `hideAlert()` would dismiss. Not a bug.

**4. (Low, test) `wrong_pin.yaml` — addressed via a test split (`7a9ccb7`), see below.**

## Test split (the main test change in this push)

The single `wrong_pin.yaml` waited for the channel-close `stillclosing` alert (`alert.button.0`) after the 10th wrong PIN. **That alert only exists when there's an open channel to close** — so on a wallet with **no** channel the wipe completes immediately, no alert appears, and the 30s `extendedWaitUntil: alert.button.0` timed out. (This also contradicted the file's own "no open channel" prerequisite.)

Split into two deterministic tests that share the 10 wrong-PIN attempts:
- **`helpers/wrong_pin_until_lockout.yaml`** — the shared 10 attempts; stops at the lockout so each caller handles its own divergence.
- **`wrong_pin.yaml`** — **no channel**: wipe completes immediately → Signup1.
- **`wrong_pin_with_channel.yaml`** — **with channel**: cooperative close → `stillclosing` → mine → "Try again" loop → Signup1.

Both **self-provision** so they run standalone: restore a wallet if none exists, and the with-channel test detects a channel (via the Receive screen's More button, only shown when `getActiveChannel() != nil`) and builds one via `helpers/ensure_bittr_channel.yaml` (`buy_signup` if needed + `buy_incoming`) if absent.

## Notes before merge
- The two code fixes haven't been compiled in CI here — needs an Xcode build. Fix #2 is trivially safe; **Fix #1's path (coop-close failure during the lockout wipe, e.g. peer offline) is hard to hit in an automated test**, so it's worth an eyeball / manual sanity check.
- The with-channel test's self-provisioning channel detection is best-effort (marked `UNVERIFIED` in the file) — the `receive.moreButton` signal and the `buy_signup` gate should be confirmed on a first real run.

PR #21, opened 2026-06-19
Add lightning/onchain payout-mode toggle on the Buy screen

RubenWaterman:
## What

Adds a per-deposit-code **payout-mode toggle** to each card on the Buy screen — a `UISwitch` (on = ⚡ lightning, off = ⛓ on-chain) with a "?" explanation — that switches how future payouts are delivered via the signed `PATCH /customer/payment-mode` endpoint. Part 2 of `customer-payment-mode-ios.md`.

## How it works

- **Read:** the existing `GET /deposit_code` refresh parses `data.payment_mode`, so the switch reflects the server's current mode on load.
- **Write:** flipping the switch signs `payment_mode:{pubkey}:{deposit_code}:{mode}:{timestamp}` with the lightning node and PATCHes it; on success it applies the **server-confirmed** mode, persists it, and reloads just that card.
- **Robustness:** the switch is disabled (with a spinner next to it) while the request is in flight, so a second flip can't race a concurrent PATCH; auto-retries once on `expired timestamp`; reverts + alerts on other errors; a node-readiness guard; failures captured in Sentry.

## Files
`CallsManager` (`.patch`), `IbanEntity.paymentMode`, `CacheManager` persistence, `BuyViewController` (PATCH + parse + reloads), `IbanCollectionViewCell` (switch + spinner + "?"), `Main.storyboard`, `Language.swift` (explanation copy), test IDs, and a `payment_mode.yaml` Maestro flow.

## Testing
`shared/flows/features/payment_mode.yaml` — self-provisioning (restores a wallet + creates an order if needed), toggles lightning↔onchain asserting the server-confirmed state, and checks the "?" explanation.

⚠️ Not compiled in CI here — needs an Xcode build. The test's switch-state assertions use Maestro's `checked` selector (drop to screenshots if unsupported).

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,10 @@ maestro test shared/flows/features/send_onchain.yaml
 maestro test shared/flows/features/send_onchain_all.yaml
 maestro test shared/flows/features/swap.yaml
 
+# Payout-mode toggle (lightning <-> onchain) on the Buy card. Self-provisioning
+# (restores a wallet + creates an order if needed), so it can run on its own:
+maestro test shared/flows/features/payment_mode.yaml
+
 # Lightning send (normal + zero-amount invoice + Lightning Address). All three
 # targets are server-side now (invoices via the e2e endpoint / request_invoice.js,
 # address = fixed e2e e2ebittr@staging.getbittr.com), so no --env is needed.
--- a/ios/bittr/Base.lproj/Main.storyboard
+++ b/ios/bittr/Base.lproj/Main.storyboard
@@ -1917,10 +1917,10 @@
                                                         <nil key="highlightedColor"/>
                                                     </label>
                                                     <collectionView multipleTouchEnabled="YES" contentMode="scaleToFill" showsHorizontalScrollIndicator="NO" showsVerticalScrollIndicator="NO" contentInsetAdjustmentBehavior="never" dataMode="prototypes" translatesAutoresizingMaskIntoConstraints="NO" id="hTZ-oO-qDM" userLabel="Iban Collection View">
-                                                        <rect key="frame" x="0.0" y="78" width="393" height="285"/>
+                                                        <rect key="frame" x="0.0" y="78" width="393" height="340"/>
                                                         <color key="backgroundColor" white="0.0" alpha="0.0" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
                                                         <constraints>
-                                                            <constraint firstAttribute="height" constant="285" id="NHM-6V-QQk"/>
+                                                            <constraint firstAttribute="height" constant="340" id="NHM-6V-QQk"/>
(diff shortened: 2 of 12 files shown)


tommelters:
@RubenWaterman  I've added the toggle now to the BuyVC with the following UI.🙏 Let me know what you think.

Simulator Screenshot - iPhone 15 - 2026-06-22 at 15 55 44
Simulator Screenshot - iPhone 15 - 2026-06-22 at 15 55 30

RubenWaterman (commented):
## Review — lightning/onchain payout-mode toggle

Reviewed the full PR end to end. The feature is complete and solid: a per-IBAN-card `UISwitch` (on = lightning, off = onchain) wired to the signed `PATCH /customer/payment-mode`, built in the storyboard (so it renders correctly, unlike the earlier programmatic attempt). Optimistic flip → signed PATCH → apply the **server-confirmed** mode + persist + reload; errors revert via reload; the 60s `expired timestamp` case auto-retries once; failures go to Sentry; a node-readiness guard + a "?" explanation round it out.

### Review findings — all addressed
- **(1) Switch race** — the switch is now disabled (with the per-card spinner) while the PATCH is in flight and re-enabled when the card reloads with the confirmed value, so a second flip can't fire a concurrent, racing request.
- **(2) Empty mode defaults to lightning** — left as-is, intentionally: orders are always created as lightning, so on/lightning is the correct default.
- **(3) Full `reloadData`** — replaced with `reloadItems` for just the affected card (new `reloadCard(for:)` helper), preserving the other cards + the horizontal scroll position.
- **(4) Two spinners** — the screen-level `updateDataSpinner` is dropped from the toggle path; the only indicator is now the spinner next to the switch.
- **(5) No test/ID** — added `buy.paymentModeSwitch` + `buy.paymentModeButton` test IDs and a `payment_mode.yaml` flow.

### Test (`features/payment_mode.yaml`)
Self-provisioning so it runs standalone: restores a wallet if none exists, creates an order via `buy_signup` if there's no deposit code (no open channel needed — switching to lightning only requires the registered node, which signup provides), then:
- toggles to onchain and back to lightning, asserting the switch settles to the **server-confirmed** state each time;
- opens the **"?" Lightning explanation** alert and asserts its copy, with a screenshot (`01b_explanation`).

Copy for the explanation was also reworded (up to 100 EUR/CHF, friendlier).

### Notes before merge
- **Not compiled here** — needs an Xcode build (changes are small/mechanical).
- The test's on/off assertions use Maestro's **`checked`** selector against the `UISwitch`; if that doesn't read switch state on iOS in your Maestro version, drop the `checked:` lines and rely on the screenshots. (Flagged in the flow header.)
- The branch-environment build script rewrites `project.pbxproj` bundle IDs to the `-regtest` variants on a build; that churn was **deliberately not committed**.

PR #23, opened 2026-06-19
New forgot_pin_remove_wallet.yaml test and associated updates

tommelters:
These updates concern the RestoreViewController:

1. User opens the app.
2. User clicks "Forgot PIN".
3. In the RestoreViewController, the user taps "Remove wallet.

**Updates**
- Since app removal may now take a while, the app now caches when the user has attempted to remove their app through the RemoveViewController.
- When relaunching the app later, the app asks immediately whether the user would like to retry app deletion. If no, the app won't ask again.
- There's now a new test forgot_pin_remove_wallet.yaml to test this flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ maestro test shared/flows/features/swap.yaml
 maestro test shared/flows/features/send_lightning.yaml
 
 maestro test shared/flows/features/remove_wallet.yaml
+maestro test shared/flows/features/forgot_pin_remove_wallet.yaml
 maestro test shared/flows/features/wrong_pin.yaml
 maestro test shared/flows/features/bitcoin_value.yaml
 maestro test shared/flows/features/bitcoin_map.yaml
--- a/ios/bittr/CacheManager.swift
+++ b/ios/bittr/CacheManager.swift
@@ -977,7 +977,25 @@ class CacheManager: NSObject {
         let defaults = UserDefaults.standard
         defaults.set(0, forKey: envKey)
     }
-    
+
+    // MARK: - Wallet removal in progress
+
+    static func setWalletRemovalInProgress(_ inProgress: Bool) {
+
+        let envKey = EnvironmentConfig.cacheKey(for: "walletremovalinprogress")
+
+        let defaults = UserDefaults.standard
+        defaults.set(inProgress, forKey: envKey)
+    }
+
+    static func walletRemovalIsInProgress() -> Bool {
+
+        let envKey = EnvironmentConfig.cacheKey(for: "walletremovalinprogress")
+
+        let defaults = UserDefaults.standard
+        return defaults.bool(forKey: envKey)
+    }
+
     // MARK: - Event handling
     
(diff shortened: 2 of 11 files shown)


RubenWaterman (commented):
## Review — resume an unfinished wallet removal on launch

Reviewed the full PR. Solid work on a sensitive path (the wallet wipe). It adds a "resume an unfinished removal on launch" mechanism, tidies the cover/spinner during a *manual* channel close, and ships a thorough test. Verified the new `removeWalletButton` test ID and the `removalinprogress` Language key are present and in sync; no compile/runtime breakers.

### What it does well
- **Cover/spinner is consistent on the manual path.** `closeChannelConfirmed` and `forceCloseChannel` raise the cover+spinner at the start and lower them on every reachable manual terminal branch (close success → alert, close throws → force-close prompt, force-close success/throw → alert, no-channel → re-gate). No stuck-cover dead-end here, and `forceCloseChannel` is only reached on the manual path so its lockout branch is dead/defensive.
- **Two resume mechanisms cleanly separated:** the incorrect-PIN lockout resumes via the `failedPinAttempts >= 10` check; the manual removal resumes via the new `walletRemovalInProgress` flag. Mutually exclusive.
- **Flag is self-healing:** cleared in `performWalletReset` (finished) and `cancelWalletRemoval` (declined).
- **Good, well-documented test** covering both branches (no channel → immediate wipe; channel → close + mine + dismiss QuestionVC + "still closing" + relaunch → resume prompt → Signup1).

### Findings + resolution
**1. (was: flag set too early) — FIXED (`1417f34`).** The flag was set in `restoreWalletTapped` as soon as a removal with an open channel was *attempted*, before the user confirmed the close — so cancelling the "close your connection?" prompt left a stale flag and the next launch wrongly offered to resume. Moved the flag-set to where the removal is actually committed/deferred (`closeChannelConfirmed`, `forceCloseChannel`, and the "still closing" branch); cancelling a confirmation prompt no longer sets it.

**2. (note — out of scope)** The lockout `closeretrylater` branch still uses a dead-end `[okay]` (the finding from PR #20). Not changed here; resolves when PR #20's "Try again" fix lands on `android`.

**3. (considered, no change)** `resumeWalletRemoval` sets `resettingPin = true` — this is **required**, not a bug: it's what lets the wipe gate (`resettingPin || removingWalletForIncorrectPin`) run `performWalletReset` on resume instead of re-prompting.

### Also added this round
Made the test **self-provisioning** (`d1f88b1`): when the device has no wallet it now creates one *with an open channel* (new `helpers/create_wallet_with_channel.yaml` = `happy_path` + `buy_incoming`), gated on the create-wallet screen being visible so an existing wallet is never wiped. So the flow runs standalone and reliably exercises the channel arc.

### Before merge
- Not compiled in CI here — needs an Xcode build.
- Manual sanity for the fix: start a removal, **cancel** the "close your connection?" prompt, kill + relaunch — you should *not* get a resume prompt (before the fix you would have).

PR #27, opened 2026-06-23
Open TransactionVC for non-Bittr incoming lightning payments

tommelters:
I've fixed an issue for incoming (non-Bittr) payments, where the TransactionVC would not open up while the ReceiveVC is still open and being dismissed.


RubenWaterman (commented):
## Review — looks correct ✅

The change fixes a real **"can't present while dismissing"** UIKit race in `launchTransactionVC`. Previously:

```swift
self.receiveVC?.dismiss(animated: true)          // starts an animated dismiss
…
self.homeVC?.performSegue("HomeToTransaction")   // presents immediately, mid-dismiss
```

When the Receive screen was open (the typical case for a non-Bittr incoming payment — you generate an invoice and someone pays it), presenting `TransactionVC` from `homeVC` while `receiveVC` was still animating out got dropped by UIKit, so `TransactionVC` never appeared. Deferring the segue into the dismiss `completion:` (and presenting directly when there's no `receiveVC`) is the right fix.

Bonus: moving `storeLightningTransaction` ahead of the segue removes a latent ordering issue where the destination could load before the transaction was cached. The `"Swap onchain to lightning"` filter behaviour is preserved and now applied consistently in both branches.

### Refinements applied (commit 38a56db)
- `[weak self]` + `guard let self` on the dismiss-completion closure, so it doesn't extend the `CoreViewController`'s lifetime across the animation.
- `if let receiveVC` instead of a `!= nil` check + force-unwrap.

No behavioural change from those.

## New: end-to-end verification via a new e2e endpoint (commit 40a24f5)

To actually *prove* this PR does what it says — rather than only manually testing it — a new regtest backend endpoint has been added:

```
POST /e2e/submit-invoice   body: { "invoice": "<bolt11>" }
```

It pays a lightning invoice from the backend's own node. `shared/flows/features/receive_invoice.yaml` now uses it (via `scripts/submit_invoice.js`) as a final step: it creates a fresh amount invoice, **stays on the Receive screen**, has the backend pay it, and asserts `transaction.yellowCard` (TransactionVC) opens. That exercises exactly the `receiveVC != nil` → deferred-segue path this PR fixes — on the old code the assertion fails (TransactionVC never shows), on the new code it passes. Confirmed passing.

Notes:
- The step needs the channel to have enough **inbound** capacity (run `buy_incoming.yaml` first); the invoice amount is kept small (1000 sats).
- The bolt11 is captured from `receive.invoiceLabel`; if that's ever truncated the submit fails loudly rather than silently.
- Endpoint is documented in `shared/docs/regtest.md`.

PR #28, opened 2026-06-26
Various app updates and Maestro test additions

tommelters:
**Updates**:
- Removed "**Bittr purchases**" and "**Bittr notification**" from Device Details and settings.yaml test.
- Removed "**bitcoin:**" from SendVC onchain address copying without amount.
- In **Transfer2VC**, when user denies push notifications, send **payment_mode** to API and allow signup to continue.
- In **BuyVC**, block payment mode **switching to lightning** if push notifications have been denied.
- Include this in buy_signup_no_notifications.yaml test.
- Include **wrong 2FA code** entry in buy_signup_no_notifications.yaml test.
- Refactor for **fiat amount String formatting**.
- Include swap status refreshing in **swap.yaml** test.
- Show **remaining pin attempts** when entering wrong pin in PinVC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,9 @@ maestro test shared/flows/features/buy_signup.yaml
 # fallback). Requires an existing wallet without a bittr account (run
 # fresh_install_skip_signup.yaml first) — it only unlocks, no auto-create.
 # launchApp's permissions config makes Maestro auto-deny the iOS notification
-# dialog; at the denied-notifications gate the test taps "Continue" to finish
-# signup on-chain, ending on Buy with the payout-mode switch OFF:
+# dialog. It enters a wrong OTP first (incorrect-code alert), then the correct
+# one; at each denied-notifications gate it taps "Continue" to finish signup
+# on-chain, ending on Buy with the payout-mode switch OFF:
 maestro test shared/flows/features/buy_signup_no_notifications.yaml
 
 # Then a feature test on the resulting wallet — opens the lightning channel:
--- a/README.md
+++ b/README.md
@@ -72,11 +72,12 @@ maestro test shared/flows/onboarding/fresh_install_skip_signup.yaml
 maestro test shared/flows/onboarding/restore_wallet.yaml
 maestro test shared/flows/features/buy_signup.yaml
 
-# Buy-signup validation + notification-gate test (unhappy path). Requires an
-# existing wallet without a bittr account (run fresh_install_skip_signup.yaml
-# first) — it only unlocks, no auto-create. launchApp's permissions config
-# makes Maestro auto-deny the iOS notification dialog so the OTP step drives
-# the denied-notification UX through to the "authorize in Settings" gate:
+# Buy-signup validation + notification-gate test (unhappy path, then on-chain
+# fallback). Requires an existing wallet without a bittr account (run
+# fresh_install_skip_signup.yaml first) — it only unlocks, no auto-create.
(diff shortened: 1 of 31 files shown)


RubenWaterman (commented):
## Review — looks good ✅

A clean multi-feature PR; reviewed each change and found no correctness bugs. It's also nicely branched off `android`, so no conflict cascade.

### App changes
- **Notifications-denied signup → on-chain** (`Transfer2ViewController` + `BuyViewController`): coherent. Declining notifications now offers Cancel/Continue; Continue registers `payment_mode=onchain` and persists it locally. Switching the payout toggle *to lightning* requires notification auth (otherwise the card reverts + an explainer shows); *to on-chain* is always allowed; the expired-timestamp retry correctly routes to `proceedWithApiCall` so it doesn't re-check. OTP auto-submit re-arms when the code drops below 6 digits. ✓
- **PIN "N attempts left"** (`PinViewController`): `10 − failed`, with the wipe returning early at 10, so it's always ≥1 — no zero/negative edge. ✓
- **Bare on-chain address copy** (`ReceiveViewController`): bare address when no amount, BIP21 URI when an amount is set (QR still encodes the URI). ✓
- **`getCorrectBitcoinValue` → `CoreViewController` + fiat-format tidy**: reuses the existing `twoDecimals()`/`toString()` helpers; compiler enforces every call site was updated. ⚠️ Intentional behavior change worth QA awareness: fiat now **rounds** to 2 decimals where the old inline code **truncated** — a confirm-screen amount can differ by 1 cent vs before. More correct, just flagging.
- **Removed Device-details "Bittr purchases" / "Bittr notification" rows**: clean — `checkBittrTransactions` was the only caller of `getBittrTransactionDetails(sendAll:)`, so dropping that path in `LoadWalletData` is justified dead-code removal. No dangling refs.

**Minor (non-blocking):** a few now-orphaned `Language` strings (`bittrpurchases`, `bittrnotification`, `bittrtransactions`, …) remain after the row removal — harmless, could be tidied later.

### Test-robustness fixes added during review (commit 45f4c42)
Two Maestro flows needed hardening to keep the suite green; folded into this branch:
- **`settings.yaml`** — fixed a flaky assertion in the Bittr-peer "Check" block: after tapping Connect, both the connect alert and the result alert carry `alert.button.0`, so waiting for it to disappear never resolved. Now waits for the Connect button (`alert.button.1`) to go away instead.
- **`swap.yaml`** — made it self-heal against the notification state this PR introduces: running `buy_signup_no_notifications.yaml` leaves the simulator's notification permission denied, which blocks swaps (`SwapManager` deviceToken check) and **can't be re-enabled via Settings**. The swap test now detects that gate at Next and reprovisions a fresh wallet with notifications allowed (`clearState` reinstall → onboarding → `buy_incoming`) before retrying, with an `assertNotVisible` guard so it fails loudly rather than silently skipping. New helpers: `helpers/swap_leg1.yaml`, `helpers/reprovision_for_swap.yaml`. (The swap recovery path needs `push_server.js` running.)

### What to run
- Headline (notifications-denied signup, wrong-2FA, blocked lightning toggle): `fresh_install_skip_signup.yaml` → `buy_signup_no_notifications.yaml`
- Device-details rows removed: `settings.yaml`
- Swap (incl. the notification self-heal): `swap.yaml`
- PIN "N attempts left": `pin_warning.yaml`

All verified on iPhone 16e / iOS 18.5.

PR #29, opened 2026-06-29
Cover the TransactionViewController in buy_incoming, buy_more and swap flows

tommelters:
- Include testing of the TransactionViewController in the existing **buy_incoming**, **buy_more**, and **swap** tests.
- Fix an issue in swap.yaml where the test would fail because of previous, unrelated unsuccessful swaps.


RubenWaterman (commented):
## Review — clean ✅

Solid, well-scoped PR that adds `TransactionViewController` coverage to the buy/swap flows. No correctness issues.

### App code (test enablement only)
- **`TransactionViewController`**: 9 new accessibility IDs, each mapped to a `TestID.Transaction.*` constant; `TestIDs.swift` and `test-ids.json` stay in sync (counts match). No logic change.
- **`HistoryTable`**: swap-row IDs become row-indexed (`history.swapPending0` / `…swapComplete0`), matching the existing `history.transactionButton0` pattern. Verified nothing references the old non-indexed IDs (the only `history.swapPending` hits left are in comments), so this breaks no flow — and it's what makes the `swapPending0` scoping correct.

### Flows
- `buy_incoming` / `buy_more` / `swap` each open the topmost transaction → `TransactionViewController` and exercise it (copy ID, fee-info alerts, add-note → Save → assert the note, swap-status sub-screen, URL button), then return cleanly. Consistent and well-commented across all three.
- The `swapPending0` scoping (only the topmost/newest swap matters) is a genuine fix and pairs correctly with the `HistoryTable` row-indexing.

### Stacked-merge note (handled on the base, not a change needed here)
This PR makes `buy_incoming` end on **Home** (it now opens the transaction and closes back). The `reprovision_for_swap.yaml` helper on the base branch (`fix/various-updates`, from the swap notification self-heal) called `buy_incoming` and then closed the *Profit* screen — the old ending — so the merge would have mis-fired `header.downButton` on Home. I've already made that helper forward-compatible on the base (`b54933d`: the close-Profit tap is now `optional: true`), so the merge into `fix/various-updates` is seamless. `swap.yaml` itself merges cleanly — the self-heal touched the first leg, this PR touches the TransactionVC asserts + `swapPending0`, in different regions.

Verified on iPhone 16e / iOS 18.5.

RubenWaterman (approved):
## Approving ✅

Reviewed in detail (see prior comment) — clean, well-scoped TransactionViewController coverage across `buy_incoming`, `buy_more`, and `swap`:

- App code is test-enablement only: 9 new accessibility IDs on `TransactionViewController` (in sync between `TestIDs.swift` and `test-ids.json`), and row-indexed swap-row IDs (`history.swapPending0` etc.) that nothing else depended on.
- The flow additions exercise the transaction screen consistently (copy ID, fee info, add-note, swap status, URL) and return cleanly.
- The `swapPending0` scoping is a real fix and pairs correctly with the `HistoryTable` change.

The one stacked-merge interaction I flagged (this PR makes `buy_incoming` end on Home, which the base's `reprovision_for_swap.yaml` didn't expect) is already handled on `fix/various-updates` (`b54933d` — that tap is now optional), so the merge is seamless. `swap.yaml` merges cleanly too.

Verified on iPhone 16e / iOS 18.5. Good to merge.

PR #31, opened 2026-06-30
Include Bitcoin QR testing in receive_onchain.yaml

tommelters:
Adds an **optional** Bitcoin QR detour to `features/receive_onchain.yaml`, inserted before revealing ten onchain addresses back-to-back.

The block is gated on the `receive.moreButton` being visible — the unified Bitcoin QR is only offered when a lightning channel exists, which is exactly when the More button renders. If there's no channel the whole block is skipped and the flow falls through to the existing onchain renew path.

### DIFF
--- a/ios/bittr/Helpers/TestIDs.swift
+++ b/ios/bittr/Helpers/TestIDs.swift
@@ -119,6 +119,8 @@ enum TestID {
         static let addressTitle = "receive.addressTitle"
         static let amountTextField = "receive.amountTextField"
         static let copyButton = "receive.copyButton"
+        static let currencyButton = "receive.currencyButton"
+        static let currencyLabel = "receive.currencyLabel"
         static let editButton = "receive.editButton"
         static let invoiceLabel = "receive.invoiceLabel"
         static let moreButton = "receive.moreButton"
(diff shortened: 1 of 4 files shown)


RubenWaterman (approved):
## Approving ✅

Clean, well-scoped, and genuinely valuable — first coverage of the unified Bitcoin QR (address + lightning invoice in one QR) end to end.

- **App code**: 2 new test IDs (`receive.currencyButton` / `receive.currencyLabel`), in sync across `TestIDs.swift` / `test-ids.json`, and actually used by the flow. No logic changes.
- **Gating is right**: the detour runs only when `receive.moreButton` is visible (a channel exists = the unified QR is offered), waiting out the QR spinner before evaluating the gate. No channel → skipped, falls through to the renew path. Both branches converge on a freshly-opened Receive, so the existing renew path works either way.
- **Assertion is the meat**: paste the unified QR → Send defaults to lightning (invoice + amount); tap Regular → the on-chain address + amount replaces it. Correctly verifies both rails. Optional alert taps + `bdkSpinner` wait handle the BDK-rescan nondeterminism.

**Minor (non-blocking):** the currency step taps "€" and asserts `EUR`, so it assumes the display currency is EUR — true in normal suite order (`settings.yaml` resets to EUR), but would fail if a prior run left CHF. Worth a one-line comment noting the assumption; fine to merge as-is.

No harmful merge interactions with the rest of the stack. Verified on iPhone 16e / iOS 18.5. Good to merge.

PR #32, opened 2026-06-30
Inspect the QuestionVC lightning channel stats in send_lightning.yaml

tommelters:
Adds a channel-inspection step to the start of `features/send_lightning.yaml`, right after the Home sync spinner settles and before the first Send.

1. Tap the balance card → MoveViewController (tapping the upper half at `50%,20%` so the profitButton overlay doesn't route to Profit, same trick as `swap.yaml`).
2. Tap the question mark next to the lightning balance → QuestionViewController.
3. Assert `question.yellowCard` **and** `question.channelView` — the channel-statistics chart, which is shown because this flow always runs with an open channel.
4. Close the QuestionViewController → Move → Home, then the existing test proceeds.

### DIFF
--- a/ios/bittr/Helpers/TestIDs.swift
+++ b/ios/bittr/Helpers/TestIDs.swift
@@ -119,6 +119,8 @@ enum TestID {
         static let addressTitle = "receive.addressTitle"
         static let amountTextField = "receive.amountTextField"
         static let copyButton = "receive.copyButton"
+        static let currencyButton = "receive.currencyButton"
+        static let currencyLabel = "receive.currencyLabel"
         static let editButton = "receive.editButton"
         static let invoiceLabel = "receive.invoiceLabel"
         static let moreButton = "receive.moreButton"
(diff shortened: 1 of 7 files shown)


RubenWaterman (approved):
## Approving ✅

Clean, well-scoped, and verified working. First coverage of the QuestionVC lightning channel-statistics view.

**What it does:** adds a channel-inspection step to the start of `send_lightning.yaml` — balance card → MoveViewController → the "?" beside the lightning balance → QuestionViewController, asserting `question.channelView` (the channel-stats chart), then closes back to Home before the existing three sends.

- **App code:** pure test-ID wiring (`move.channelButton`, `question.channelView`) on `MoveViewController` / `QuestionViewController`, in sync across `TestIDs.swift` / `test-ids.json`, and both used by the flow. No logic changes.
- **Correctness:** the open-channel assumption is sound — `question.channelView` only renders with a channel, which `send_lightning` already requires, so it fails early on inspection rather than mid-send. Uses the `50%,20%` balance-card trick (consistent with `swap.yaml`); screenshots slot in as `01a`/`01b` with no collisions.

**Non-blocking note:** this PR's diff redundantly includes #31's Bitcoin-QR commit (`968a21b1`) — I confirmed it's byte-identical to #31's `receive_onchain.yaml`, so it's conflict-free, but a rebase onto current `android` would make the diff just the channel-stats change. Fine to merge as-is.

Verified on iPhone 16e / iOS 18.5. Good to merge.

PR #33, opened 2026-07-03
Fix some gaps in test coverage

tommelters:
- Audit of missing tests, updated in screens.md and parity.md
- Add testing of SendVC question marks in send_onchain.yaml.
- Add testing of HomeVC syncStatusView and long press on ReceiveVC QR in receive_onchain.yaml.
- New test send_swap_suggestion.yaml to test SendVC swap suggestion when lightning balance is insufficient.
- In SendVC, parse and show invoice amount even if balance is insufficient.
- In SwapVC, during suggested onchain-to-lightning swap, correctly update onchain balance.
- In SwapVC and SendVC, handle event in which BDK is unavailable, and show alert.


RubenWaterman (approved):
## Approving ✅ — strong PR, with a few improvements folded in during review

Closes real test-coverage gaps and fixes genuine behavior along the way. Reviewed the app logic, flows, and docs.

### App logic (verified) ✅
- **Real bug fix — `ReceiveViewController`:** `qrImageView.isUserInteractionEnabled = true`. `UIImageView` defaults to `false`, so the QR long-press Copy/Share context menu never received touches — this makes it work (and the new `receive_onchain` step covers it).
- **BDK-failure UX:** the previously-silent "Could not scan/start BDK" paths now surface a user alert (`presentOnchainSyncFailedAlert` → "Syncing issue"). Verified manually by taking electrum offline — the alert fires correctly from both Send (Regular/on-chain) and Swap (on-chain→lightning, incl. the swap-suggestion path).
- **BDK-gated suggested swap** (`startSuggestedOnchainToLightningSwap` / `awaitBdkScan`): waits for BDK to be ready before swapping, or aborts with the failure alert and resets the Next button. Fixes the "on-chain balance not updated during suggested swap" issue.
- **`AddressParsing`:** now shows the parsed invoice amount even when it exceeds lightning capacity (deferring insufficient-balance handling to Next's swap suggestion); the unified-QR onchain fallback is preserved.

### Improvements folded in during review
1. **DRY the failure alert (`9ee8bf8`).** `presentOnchainSyncFailedAlert()` was duplicated verbatim in `SendViewController` and `SwapViewController`. Extracted into a small `OnchainSyncFailureReporting` protocol (constrained to `UIViewController`) with a default implementation in `AlertManager.swift`; both VCs conform via their existing `bdkSpinner` outlet, call sites unchanged. One definition instead of two. Confirmed builds + both alerts still fire (electrum-offline test).

2. **Bound the BDK sync waits against a *hung* electrum (`cb95954`).** Tracing the sync path showed the original `awaitBdkScan` poll (and, more fundamentally, the blocking `fullScan`) has no timeout: a *refused* electrum fails gracefully (fullScan throws → alert), but a *hung/black-holed* electrum (accepts the connection, never responds) leaves `bdkWalletIsScanning` stuck true and every caller on a spinner forever.
   - `BDKManager.didSyncBdkWallet`: a 180s watchdog fails the sync once (on main, clears the flag) so **all** callers recover via the alert instead of hanging. Existing completion sites route through a shadowed once-guarded closure, so they're unchanged.
   - `SwapVC.awaitBdkScan`: its poll now respects the same deadline.
   - **⚠️ Tuning note:** the 180s ceiling sits above a legitimately slow scan (documented "seconds to minutes") — safe on regtest (scans are seconds), but the mainnet value should be revisited before release so a genuinely slow sync isn't reported as a failure.

### Test coverage
New `send_swap_suggestion.yaml` exercises the swap-suggestion logic; `send_onchain` (info "?" marks) and `receive_onchain` (QR context menu + sync status) cover the rest. Note `send_swap_suggestion.yaml` needs the special *on-chain-funds-but-no-channel* wallet state (documented in its header) — the one test here you can't run from a fresh install.

Verified on iPhone 16e / iOS 18.5. Good to merge (revisit the mainnet sync-timeout value as a follow-up).

PR #34, opened 2026-07-06
New send_swap_suggestion_onchain.yaml test

tommelters:
- **Suggested swaps** (Swap&Pay in the SendVC) would stay forever "pending" in the HomeViewController, because there's no incoming transaction. This is now fixed in the HomeViewController and the TransactionViewController.
- In the TransactionViewController, correctly distinguish the total paid fees of suggested swaps.
- Renamed existing test send_swap_suggestion.yaml to **send_swap_suggestion_lightning.yaml** (paste an invoice into the SendVC, then Swap&Pay).
- New test **send_swap_suggestion_onchain.yaml** (paste an onchain address into the SendVC, then Swap&Pay).
- Stop TransactionVC from popping up during suggested lightning-to-onchain swap (matching behavior for all other swaps).


RubenWaterman (commented):
Reviewed the full diff and pushed three follow-up commits addressing the findings. Summary below.

## What the PR does well

- The suggested-swap plumbing checks out end-to-end: `isSuggested` is persisted in the swap file, the `suggestedswaps` marker keeps Home from rendering the payment as a perpetually pending swap, and the TransactionVC branches (amount from the swap file, spread+network fees, direction-aware ID stacks) mirror the SwapStatus screen correctly.
- `send_swap_suggestion_onchain.yaml` is well-constructed: documents its non-provisionable prerequisite (< 60 000 sats onchain, > 60 000 outbound) instead of pretending to self-provision, handles the Allow Paste prompt, the optional syncing alert, and the SendToSwap dismissal path. README/parity.md/screens.md all consistent with the rename.
- `<b>` amounts render correctly via `attributed()`, and Language.swift is single-locale so no other dictionaries were missed.

## Findings fixed in follow-up commits (pushed to this branch)

**1. "Succeeded" was marked at broadcast time, not completion time** → fixed in 6669281.
`storeSuggestedSwap` ran right after the lockup broadcast (onchain→lightning) / at reverse-swap creation (lightning→onchain), and LoadWalletData stamped every marked transaction `.succeeded`. If Boltz then failed to pay the invoice, Home showed a succeeded outbound payment the recipient never got until the refund merge caught up. The marker now stores pending/succeeded/failed: success is recorded where the recipient is actually paid (claim broadcast for lightning→onchain; invoice.paid/transaction.claimed etc. via SwapStatusVC for onchain→lightning), failures are recorded on the terminal failure statuses, and legacy Bool markers are treated as succeeded so existing wallets don't regress.

**2. Swap payments were identified by description substring matching** → fixed in 3dc7a73.
Three sites suppressed the auto-opening TransactionVC by matching `lnDescription.contains("Swap onchain to lightning ")` (one of them had already drifted to checking only one direction). Severity note: this was *not* third-party exploitable — `lnDescription` comes from the app's own description cache, not the paid invoice's bolt11 description — but it was fragile against self-inflicted collisions and future cache writers. Replaced with a computed `Transaction.isSwapPayment` that checks whether the description keys into the `swapids` dictionary, which only the swap flows write. `storeSwapID` always runs before the corresponding payment event fires, so the lookup is populated in time.

**3. Missing sentence-ending period** in `lightninginsufficientfunds` → fixed in 833b7ae. The BOLT12 and both-sides-insufficient alerts ended mid-sentence; its three sibling strings had kept their periods.

## Remaining (optional fast-follows, not blocking)

- TransactionViewController loads the same swap file three times per render (amount, fees, IDs) — could be loaded once.
- LoadWalletData duplicates the swap-direction detection (`contains("onchain to lightning")`) 20 lines apart.
- The `suggestedswaps` UserDefaults dictionary grows unboundedly, but so do its siblings (`swapids`, descriptions) — fine for now.

Recommend re-running both suggested-swap maestro flows before merge, since commits 6669281/3dc7a73 touch exactly the paths they exercise (marker timing around "Swap complete", and the segue suppression in the lightning variant).

PR #35, opened 2026-07-06
New tests for push notifications

tommelters:
- New test **notification_information.yaml** now tests push notifications with type .information, .htlcExpired, .unknown.
- New test **notification_lnurl.yaml** now tests the .lnUrl push notification.
- New test **notification_htlcincoming.yaml** now tests the .htlcIncoming push notification.
- I've replaced the "receiving payment" pendingView with a cleaner solution, a card sliding up from the bottom of the screen, similar to the custom alerts.


RubenWaterman (commented):
Reviewed the diff (the new push-notification work — the swap-related files in this diff are inherited from the stacked #34 branch and were reviewed there). Pushed three follow-up commits addressing the smaller findings; details below.

## What the PR does well

- The three new flows close a real coverage gap: five of the six `BittrNotificationType` cases now have APNS-injection flows (only `.swap` remains, and parity.md correctly says so). Each flow documents its state machine (locked vs synced vs deferred), why the terminal failure alert is the right assertion without real backend state, and the notification-banner-swallows-the-tap workaround.
- The dedup-aware re-push loops are sound against the code: `handleNotification` only resets its 10s window for notifications it *accepts*, so swallowed re-pushes don't extend the window and the loops terminate.
- The `showLoading`/`hideLoading` overlay is a genuine upgrade over the fixed storyboard pendingView, faithfully mirrors `showAlert`'s technique, and no stale `pendingView`/`showPendingView` references remain. All call sites pair show/hide on the same VC.
- `htlcReadyFriendlyMessage` is a clean seam: raw backend codes (`no_held_htlc`) no longer reach users, the catch path correctly unwraps `BittrServiceError.serverError`, and the raw code is still logged for debugging.
- README / flows README / parity.md / screens.md all updated consistently.

## Fixed in follow-up commits (pushed to this branch)

- **`paymentrequest3` never substituted `<amount>`** → 87a4453. The alert was shown with no `replacingOccurrences` call, and since `attributed()` HTML-parses messages, the literal `<amount>` tag was silently dropped — the signed-in payment-request alert read "Someone wants to pay you satoshis!" with no number (with a bolded orphan "satoshis" after this PR's `<b>` markup). Now substitutes like the signed-out site; also added `.addSpaces()` to both sites for the space-grouping convention.
- **First push in `notification_information.yaml` wasn't dedup-protected** → 21be5f9. Blocks 2/3 re-push until the card opens, but block 1 pushed once — a notification handled by an earlier flow in a suite run within 10s would swallow it. Now uses the same repeat-until-visible pattern.
- **Transient loading-card waits raced the sync** → e839f26. `notification_lnurl.yaml` failed a real run on `Assert ".*[Ss]yncing wallet.*"` even though the card visibly appeared: on a warm wallet the sync finishes between two Maestro hierarchy polls. The intermediate "syncing wallet" / "receiving payment" checks in both locked-push flows are now `optional: true` (the settings.yaml pattern); the terminal alerts — which wait for a tap and can't be missed — remain the hard assertions.

## Open findings (for discussion, not fixed)

1. **The loading overlay hard-locks the app if any path misses `hideLoading()`.** The card is non-dismissable and the full-screen background absorbs every touch, with no timeout or escape hatch. All current paths look correctly paired, but the pairing spans async network callbacks in `facilitateHTLCReady`/`facilitateNotificationPayout` — one future early-return without `hideLoading()` bricks the session until relaunch. Consider a defensive auto-dismiss (60–90s) or a hold-to-dismiss escape.
2. **Stacked-branch note**: this branch predates the three review-fix commits now on `fix/suggested-swaps` (6669281, 3dc7a73, 833b7ae). I checked hunk overlap — no conflicts expected in either merge order — but rebasing onto the updated `fix/suggested-swaps` before merge would let these flows run against the final swap code.

## Nits

- README's "needs push_server" sentence names `buy_more`, `notification_information` and `notification_lnurl` but omits `notification_htlcincoming` (its run-line comment does say it).
- `QuestionViewController` `channelStackHeight` 30→20 is unexplained in the PR body (harmless layout tweak, just worth a mention).
- `sleep.js` busy-waits the host CPU — fine given Maestro's GraalJS has no sleep, and the comment says why.

Recommend a run of `buy_more.yaml` before merge as a regression check: it exercises the `.lightningPayout` path whose pendingView was replaced by the new overlay.

PR #36, opened 2026-07-07
New test fresh_install_unhappy.yaml

tommelters:
- New test **fresh_install_unhappy.yaml** to test all possible alerts during wallet creation.
- Requires to run **node shared/flows/scripts/screenshot_server.js** to take screenshots.
- Requires that **Terminal has Accessibility access** (Apple icon > System Settings > Privacy & Security > Accessibility > + > Applications > Utilities > Terminal.
- Requires **on the simulator**, go to Settings > General > Screen Capture > **Full-Screen Previews OFF**.
- Small guard for listenForEvents() crash.
- Small fix for PinVC numbers highlighting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,16 @@ maestro test shared/flows/onboarding/fresh_install.yaml
 # (taps "Skip" on Signup7 and lands on Home):
 maestro test shared/flows/onboarding/fresh_install_skip_signup.yaml
 
+# Option 1c: Full reset + wallet creation through every validation gate
+# (confirm-statements / screenshot-warning / invalid-word / wrong-phrase /
+# PIN-length alerts), then Skip to Home — the onboarding error-path counterpart.
+# To exercise the seed-phrase screenshot warning (optional): run
+# "node shared/flows/scripts/screenshot_server.js" (needs Accessibility
+# permission), AND on the simulated device turn OFF Settings > General > Screen
+# Capture > Full-Screen Previews (else the screenshot preview covers the app and
+# blocks the flow). Without this setup the screenshot step is skipped.
+maestro test shared/flows/onboarding/fresh_install_unhappy.yaml
+
 # Option 2: Full reset + restore existing wallet.
 # Followed by onboarding through the Buy page.
 maestro test shared/flows/onboarding/restore_wallet.yaml
--- a/ios/bittr/BitcoinManager.swift
+++ b/ios/bittr/BitcoinManager.swift
@@ -605,6 +605,7 @@ class BitcoinManager {
             guard let self else { return }
             
             while !Task.isCancelled {
+                guard self.ldkNode != nil else { return }
                 let event = await self.ldkNode!.nextEventAsync()
                 if Task.isCancelled { break }
                 await MainActor.run {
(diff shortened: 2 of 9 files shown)


RubenWaterman (commented):
Reviewed the PR's own two commits (db4ae25, 6be4f4c) — the rest of the GitHub diff is inherited from the stacked #34/#35 branches, reviewed there.

## What the PR does well

- **fresh_install_unhappy.yaml** covers every onboarding error branch — Cancel from the confirm screen, confirm-statements alert, seed-phrase screenshot warning, invalid non-BIP39 word, valid-but-wrong phrase, and PIN too-short / too-long / mismatch — and every asserted alert text matches a real Language.swift string (verified: signupvc2, noscreenshot, invalidwords2, incorrectphrase2, pinshouldbe4to8, pincanbeupto8, repeatnumber).
- **screenshot_server.js / trigger_screenshot.js** is a clever, well-documented solution to a genuine gap: Maestro's takeScreenshot never posts `userDidTakeScreenshotNotification`, so the helper AppleScript-clicks the Simulator's "Device > Trigger Screenshot", which does. Correctly optional end-to-end (helper unreachable → flow still passes), localhost-only, fixed script (no injection surface), no port collision (8890 vs 8888/8889). Both host-side prerequisites (Accessibility, Full-Screen Previews OFF) are documented with the why.
- **PIN highlight fix** is correct: the touch-down handler lights the button, and the delayed reset previously sat after the max-length early return — so a 9th-digit tap left the button stuck lit behind the alert. Moving the reset to the top covers every exit path.
- Docs (README, flows README, parity.md, screens.md) all updated consistently.

## Findings (acknowledged — to be addressed in a follow-up PR)

1. **The listenForEvents guard narrows the crash window but doesn't close it** (BitcoinManager.swift:608). It's check-then-force-unwrap: `resetNodeState` can nil `ldkNode` while the loop is suspended inside `nextEventAsync()`, after which the unguarded `self.ldkNode!.eventHandled()` three lines down hits the same crash. Binding once per iteration (`guard let node = self.ldkNode else { return }` and using `node` for both calls) fixes all three uses. The `return` itself is fine — FetchAndPrint restarts the listener on each sync.
2. **The screenshot-warning capture can silently skip even with the helper running**: trigger_screenshot.js returns as soon as the menu item is clicked, but the alert's slide-up takes a beat, and the `runFlow when: visible` check evaluates immediately — if it polls first, the optional block skips with no signal. A short pause between runScript and the runFlow (sleep.js ~1s or waitForAnimationToEnd) would make the 06_screenshot_warning capture reliable.

## Nits

- The twelve copyTextFrom/evalScript pairs are bulky but idiomatic — Maestro has no loop-over-dynamic-ids, and it matches happy_path_wallet.yaml's pattern.
- The flow ends with a fresh wallet (PIN 1234), consistent with the happy path, so downstream suite flows can reuse the state. Worth remembering it clearState+clearKeychain wipes whatever wallet the simulator had.

Branch logistics: the stacked-branch conflicts against android (same squash-merge cause as #35 had) are being resolved by rebasing onto android now that #35 is merged.

PR #37, opened 2026-07-07
Filling additional test gaps from parity.md

tommelters:
- Include Transfer1VC "I don't have an IBAN" in fresh_install_unhappy.yaml.
- Include Transfer2VC "Resend code" in buy_signup_no_notifications.yaml.
- Include Transfer3VC copying and "Screenshot" in buy_signup_no_notifications.yaml.
- Include Transfer4VC "Back" button in buy_signup_no_notifications.yaml.
- Add drag-to-dismiss to OnePlaceVC popup in MapVC.
- Include OnePlaceVC website button and "Open in Maps" in bitcoin_map.yaml.
- Include OneLessonVC Back button in academy.yaml.
- Improve SendVC LNURL spinner UI.


RubenWaterman (commented):
Reviewed the PR's own six commits (the GitHub diff previously re-showed the stacked #34–#36 work; after the rebase it's exactly the 20-file own scope). Conflicts against android are already resolved — the branch was rebased onto android post-#36-merge and is mergeable.

## What the PR does well

- Disciplined parity.md gap-filling: Transfer1 "I don't have an IBAN" exit, Transfer2 resend + cooldown + Change-email round-trip, Transfer3 copy IBAN/name/code + Screenshot-to-Photos, Transfer4 Back, the map place website + "Open in Maps" round-trip, and the Academy lesson Back button — each with accurate assertions (all verified against Language.swift: "Handling LNURL", resendcode2, onlyiban) and screenshots.
- The drag-to-dismiss on the map's OnePlace popup is well implemented: downward-only clamp, dismiss on distance (>1/3) or velocity (>800), spring-back otherwise, dim fade tracking the drag, and isModalInPresentation toggling so the popup drag doesn't fight the map sheet's own pull-to-dismiss.
- Consolidating the Send LNURL spinner onto the shared showLoading overlay removes a storyboard view + 4 outlets + a test ID with zero stale references left, and send_lightning.yaml is correctly re-anchored to the overlay's label text (plain text, so the matcher works).
- All new test IDs are mirrored in both TestIDs.swift and shared/test-ids/test-ids.json.

## Findings (non-blocking)

1. **The Apple Maps return trip is the one fragile step** (bitcoin_map.yaml): a fixed-coordinate tap (13%,4%) on the SpringBoard "‹ bittr regtest" breadcrumb — pinned to iPhone 15, the app display name, and Apple Maps foregrounding without first-launch sheets covering that spot. The comment documents all this honestly. Sturdier alternative worth trying: `- launchApp:` with `stopApp: false`, which foregrounds Bittr directly without coordinates or SpringBoard elements and keeps app state. If it misbehaves, the documented tap is an acceptable fallback.
2. **The LNURL path now inherits the loading overlay's no-escape-hatch issue** (deferred from #35's review): startLNURLSpinner shows the non-dismissable overlay during network resolution, so a hung LNURL endpoint leaves Send bricked until relaunch. Not a regression — the old spinner view was equally full-screen and blocking — but the trigger is now a user-pasted address, which strengthens the case for the defensive auto-dismiss when that follow-up PR happens.

## Nits

- The Photos permission handling (runFlow when visible "Allow") is correct — Maestro's full-string match can't hit "Don't Allow" — and self-documenting about iOS-version text drift.
- The resend block's 5s sleep sits safely inside the 30s cooldown, and re-entering credentials after "Change email" defensively erases first.
- academy.yaml's back-forward dance waits out the image spinner between every step, matching the file's existing gating idiom.

PR #38, opened 2026-07-13
Minor test improvements and bittrWallet refactor

tommelters:
- **QuestionVC** now has bold text when explaining the lightning channel statistics.
- Refactor bringing **bittrWallet** from CoreVC into BitcoinManager. Cleanup of associated code.
- In **suite.yaml**, move swap.yaml before send_onchain.yaml, so that the wallet has onchain satoshis.
- In **receive.yaml** and **restore_wallet.yaml**, wait for complete wallet sync.
- **Cache funding transaction** upon channelReady event, so that it's immediately visible upon wallet relaunch.


RubenWaterman (commented):
Reviewed the PR's own seven commits. Conflicts against android were resolved by rebasing onto android post-#37 (the usual squash-merge stacking); the PR now shows its own 44-file scope and is mergeable. One follow-up commit pushed (d309837, see finding 1).

## What the PR does well

- **The bittrWallet refactor lands clean — verified, not assumed.** Zero `coreVC…bittrWallet` references remain anywhere, including in code merged to android *after* this branch forked (e.g. SendOnchain.swift:63 now carries both the refactor's `BitcoinManager.shared.bittrWallet` and the space-grouping fix from #34's review). Moving wallet state off a view controller onto the singleton that owns the node is the right direction.
- **The relaunch-persistence fix addresses a real observed bug** (missing funding tx in buy_incoming, missing 2nd payment in buy_more): `addLightningTransaction` now persists `visibleTransactions` through the existing `"transactions"` serialization path, so `ShowCachedData`'s restore works unchanged. The cache-write-before-duplicate-guard ordering is deliberate and correct — the cache always mirrors the table while balance math stays gated to non-duplicates; the commit message explains it well.
- The flow tunings all target known flake sources: sync-completion waits in receive/restore_wallet (so suite successors start from a settled wallet), the share-sheet animation wait, and the self-healing swipe-down fallback for the Send dismiss (nice `runFlow when: notVisible` idiom).
- questionvc1/questionvc7 gained matching `<b>` markup; the no-channel fallback strings render fine through `attributed()`.

## Findings

1. **[fixed in d309837]** The refactor accidentally deleted the 11-line doc comment on `channelsFullyClosedAndSwept()` — the one explaining that wiping before channels are swept destroys the LDK channel monitors (unreconstructable from seed; CSV-delayed force-close outputs) and means permanent fund loss. Restored verbatim on this branch.
2. **`attributed()` centers text — please eyeball the QuestionVC alignment.** The helper wraps content in `<center>`, so the three channel-stat answers (and no-channel fallbacks) now render centered if `answerLabel` was left-aligned before — while other question types still use plain `.text`, which would leave alignment inconsistent across questions. If centering is wrong, `attributed()` deserves an alignment parameter rather than per-call workarounds.
3. **Observation, no action needed now**: `bittrWallet` was already mutated from both main-thread UI code and background sync tasks on CoreVC; that's unchanged on the singleton. The refactor makes an eventual serialization (actor/queue) easier — worth keeping on the radar.

## Nits

- `addLightningTransaction` now writes the full serialized transaction list to UserDefaults on every live event including duplicate re-deliveries — fine at wallet scale, just noting it's the hot path during payment bursts.
- The suite.yaml reorder comment says what but not why — worth carrying the PR body's "so the wallet has onchain satoshis" rationale into the yaml comment.

## Verification

`suite.yaml` end-to-end is the real test since the refactor touches everything; plus buy_incoming/buy_more specifically for the relaunch-persistence fix (receive while on Home → relaunch → the tx should be in the table immediately), and send_onchain's "You can send…" question for the bold stats.

PR #39, opened 2026-07-13
suite.yaml and swap testing improvements, new custom textfield alert

tommelters:
- **New custom textfield alert** in **TransactionVC** when adding a note.
- **Bold Bittr details** in **Transfer4VC** in the "Create a recurring transfer" alert.
- **suite.yaml** now starts with **fresh_install_skip_signup.yaml** instead of restore_wallet.yaml (because the restored wallet doesn't sync properly). 
- In **suite.yaml** I've moved **restore_wallet.yaml** down before forgot_pin.yaml.
- In **suite.yaml** I've moved **send_onchain_all.yaml** up directly after **send_onchain.yaml**, so that the onchain wallet is empty ahead of the swap suggestion tests.
- I've included the **send_swap_suggestion_onchain/lightning.yaml** tests in **suite.yaml**, directly after send_lightning.yaml.
- In **send_swap_suggestion_onchain.yaml**, the test now tries to send satoshis to an onchain address from its own wallet (leading to a suggested swap). The amount of this transaction is calculated dynamically during the test, in order to empty the lightning connection. So the test ends with a full onchain wallet and an empty lightning connection, ahead of the next send_swap_suggestion_lightning.yaml test.
- Additional fixes in **SwapViewController** for loading the BDK sendable amount and showing the spinner.


RubenWaterman (commented):
Reviewed the PR's own three commits, plus this branch accumulated four hardening commits during suite testing (details below). Conflicts against android were resolved early on by rebasing onto android post-#38; the PR shows its own scope and is mergeable.

## What the PR does well

- **showTextFieldAlert** is a careful house-style replacement for the native UIAlertController: same card and slide-up motion, `alert.button.<n>` + `alert.textField` test IDs, keyboard avoidance, and a keyboard-aware two-phase dismissal with real motion-quality work (observers torn down first so will-hide doesn't fight the animation; `endEditing`'s return value distinguishes keyboard-up from already-down — and the comment explains why hideAlert's off-screen math needs the card at rest). Observers are also torn down defensively in hideAlert, so nothing leaks if something else dismisses the alert. All three note-adding flows updated.
- **send_swap_suggestion_onchain.yaml becomes self-contained**: live-balance sizing (`send = 50000 + max(0, instant − 75000)` with documented floor math), pays the wallet's OWN address, drops the clipboard_server dependency, and deliberately ends in the "full onchain, empty lightning" state the next suite flow needs.
- **The suite reorder fixes a real latent bug**: with the suite now starting from a fresh-install random seed, forgot_pin's MNEMONIC env could never match mid-suite — moving it after the late restore_wallet (which re-establishes the fixed-mnemonic wallet) is correct, and the comment explains the dependency.
- **SwapVC sendable-amount**: spinner during fee/PSBT computation plus a captured requestedDirection guard against stale async label updates.
- Transfer4's recurring-transfer alert renders the IBAN/name/code bold via placeholders; `bankingapp2`'s mixed `<br>`/`\n` markup renders correctly through attributed().

## Findings (non-blocking)

1. **The suite's opening switch is a workaround with no tracker**: "the restored wallet doesn't sync properly" is real enough to restructure the suite around, but the underlying restore-sync bug remains — worth an issue so the green suite doesn't hide it.
2. **The sweep amount has no upper bound**: fine inside the suite (prior flows bound the lightning balance), but a standalone run against a fat channel could exceed Boltz's max swap size and fail with an unrelated-looking error. A one-line cap or a comment stating the assumed range would help.
3. The direction guard accepts one theoretical stale case (flip A→B→A fast enough and an in-flight first-A response still passes). Harmless — noting the guard is per-direction, not per-request.

## Hardening commits added to this branch during suite testing

- **9f15153 — retry transient 5xx in the staging-facing Maestro helpers.** A full suite run died on a single 502 from trigger_bank_transaction.js. All four staging helpers now retry up to 5× with linear backoff on network errors/5xx (4xx still fails fast); per-endpoint idempotency reasoning is in the commit message.
- **1a435a4 — positive-evidence provisioning gates in the wrong_pin flows.** A gate on "unlock screen not visible" evaluated one second after launchApp — during the launch animation — concluded "no wallet" and clearState-wiped the suite's live channel wallet mid-run. Both flows now wait for the animation to settle and restore only when the create-wallet screen is actually visible (the pattern payment_mode.yaml already documents). Evaluated too early, the positive gate merely skips and fails loudly; it can never destroy a good wallet.
- **5df37e8 — BlurEffect animator pile-up hard-hang.** Sentry AppHang reports (main thread fully blocked in setFractionComplete → UIVisualEffectView deferred animations) traced to the Academy locked-lesson blur: animation blocks accumulated on a reused animator, every live instance responded to each broadly-posted "setupblur", and detached instances never unregistered. The watchdog kill + relaunch is why suite runs ended on a relaunched PIN screen instead of Signup1. Fixed: animator recreated per call, detached instances skip and tear down, no duplicate observers.
- **b151e7d — listenForEvents nil-node race** (deferred from #36's review, made current by wallet-reset crashes): the loop now binds the node once per iteration, so resetNodeState nil-ing ldkNode mid-nextEventAsync can't crash eventHandled.

## Verification

Full `shared/flows/test_suite.sh` (rebuild the app first — two of the hardening commits are app code). Targeted: buy_incoming/buy_more for the note alert, swap.yaml for the pending-swap recovery, the two suggested-swap flows back-to-back in suite order, and a menu-tap into Academy after scrolling (the hang trigger) for 5df37e8.

PR #40, opened 2026-07-13
Refactoring and improvements to background thread usage

tommelters:
- Improvements to **CoreVC** launch, showing either SignupVC or PinVC.
- Refactoring of **BitcoinManager** and **BDKManager** functions that were **redundantly asynchronous**.
- Auditing and refactoring of **background thread usage** for functions that shouldn't block the main thread.
- Minor cleanup of performSwapMatching function.

### DIFF
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -11,92 +11,76 @@ import Sentry
 
 extension BitcoinManager {
     
-    func didStartBDK(completion: @escaping (Bool) -> Void) {
-        DispatchQueue.global(qos: .background).async {
-            
-            // BDK launch.
-            if self.bdkWallet == nil {
-                Log.info("Will start blockchain and wallet.")
-                
-                // Bail out if the mnemonic has been cleared from the cache. This
-                // happens during a wallet wipe: performWalletReset deletes the
-                // mnemonic (deleteClientInfo) and clears bdkWallet (resetNodeState),
-                // so a didStartBDK that runs after teardown would force-unwrap a
-                // nil mnemonic and trap with a Swift _assertionFailure.
-                guard let cachedMnemonic = CacheManager.getMnemonic() else {
-                    Log.info("No mnemonic in cache; wallet is being torn down. Aborting BDK start.")
-                    completion(false)
-                    return
-                }
-
-                // Attempt to create a mnemonic object from the provided mnemonic string.
-                let mnemonic:BitcoinDevKit.Mnemonic
-                do {
-                    mnemonic = try BitcoinDevKit.Mnemonic.fromString(mnemonic: cachedMnemonic)
-                } catch {
-                    self.handleError(error: error, row: 178)
-                    completion(false)
-                    return
-                }
-                
-                // Create a BIP32 extended root key using the mnemonic and a nil password
-                let bip32ExtendedRootKey = DescriptorSecretKey(network: EnvironmentConfig.bitcoinDevKitNetwork, mnemonic: mnemonic, password: nil)
-                
-                // Create a BIP84 external descriptor using the BIP32 extended root key, specifying the keychain as external and the network as testnet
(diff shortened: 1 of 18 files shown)


RubenWaterman (commented):
Reviewed the PR's own five commits. Conflicts were resolved by rebasing onto android post-#39 — one real conflict in performSwapMatching (the cleanup was written against pre-#34 code and would have regressed the suggested-swap status back to hardcoded .succeeded; resolved keeping the honest status plus the tidier ternary). Four follow-up commits are pushed to this branch; details below.

## What the PR does well

- **The reset-path background moves are genuinely valuable**: performWalletReset's node-stop/document-deletion, the channel-close paths' listChannels+closeChannel, and both syncWallets call sites were real main-thread stalls (stopping the LDK node on main was very plausibly a contributor to the reset-time freezes we chased this week). syncLDKnode's restructure also fixes a latent cross-thread write — lightningChannels is now assigned on main.
- **De-asyncing didStartBDK/didStartLDK is the right call structurally** — the completion-based versions were needlessly callback-shaped for sequential logic, and the new call sites read much better.
- The PIN-entry flattening preserves the lockout-before-comparison semantics exactly while being far more readable, and the observer/test-ID/gesture setup relocation lands in setBasicStyling (called once from viewDidLoad — verified it's not in the repeatedly-invoked setWords, so no duplicate registrations).

## Findings — all fixed in follow-up commits on this branch

1. **Build break** → e544d31. AnimationViewController still called the deleted showPinOrSignup(); the call-site update lives in the fix/launch-animation branch, so this PR alone didn't compile (verified on the original branch tip too — not a rebase artifact). Now calls checkWalletRemoval(), matching that branch's intent.

2. **Suite-breaking launch regression** → e544d31. Moving the signup/pin container alphas to viewDidLoad put their buttons into the accessibility tree while the opaque launch-animation cover still owned all touches — Maestro's visibility gates opened seconds early and the first tap was silently swallowed by the cover. This broke the whole suite at its first flow on every clearState launch (warm launches sometimes won the race, hence the flaky look). Diagnosed live: tap "succeeds", main thread idle, nothing presents; a second tap after the animation presents instantly. Fixed by keeping the refactor's shape but deferring the reveal: new revealPinOrSignup(), called from the animation as its cover drops. Verified twice against a rebuilt app with the previously-deterministic clearState repro.

3. **UIKit from a background thread** → a54ef74. continueStartWallet called loadWalletData() directly on the userInitiated queue; it does direct UI work (labels, table reload, finalizeSync, alerts). The sequencing intent of the restructure is preserved — the call is hopped to main, same as the hop syncLDKnode kept. Splitting loadWalletData into data/UI phases is the better follow-up.

4. **The synchronous didStartBDK() blocked main at three call sites** → a54ef74. Both bdkWalletUnavailable() implementations and the suggested-swap BDK preparation run on main, and didStartBDK does descriptor derivation + SQLite + electrum client setup — the old completion version hid a background dispatch inside, which is why those sites were safe. Each now hops the call off main and marshals UI outcomes back.

5. **The 15s LDK-start watchdog was dropped** → a54ef74. A hung node build/start meant spinner-forever; restored with the same deadline and retry-alert semantics the old task-group race had.

6. **Comment casualty** → 6b0680f. The performSwapMatching cleanup deleted the explanation of why summing per-leg fees is safe (inbound legs have fee = 0) — restored, same category as #38's channelsFullyClosedAndSwept comment.

Also on the branch: e5184ae makes test_suite.sh checkpoint per-flow progress so failed suite runs resume instead of starting over (suite.yaml stays the canonical flow list).

## Verification

Full suite via the new resumable test_suite.sh (rebuild first — several fixes are app code). The clearState-launch repro for finding 2 passes twice; findings 3–4 need Main Thread Checker eyes in Xcode for extra confidence. Separately noted for a future PR: a pre-existing force-unwrap crash in checkOnchainAddressesWithBittr surfaced in today's crash logs (not caused by this PR).

PR #41, opened 2026-07-14
Minor updates to AnimationVC and AcademyVC

tommelters:
- Refactor to AnimationVC launch animation, using iOS usingSpringWithDamping animation for a smoother animation.
- Make sure PinVC and SignupVC are tappable through the AnimationVC, as not to block them while they become visible.
- Fix dark mode issue for AcademyVC blur view being too dark.

### DIFF
--- a/ios/bittr/Academy/Academy/AcademyVC/LessonCollectionViewCell.swift
+++ b/ios/bittr/Academy/Academy/AcademyVC/LessonCollectionViewCell.swift
@@ -81,6 +81,19 @@ class BlurEffect: UIVisualEffectView {
 
     var blurAnimator = UIViewPropertyAnimator(duration: 1, curve: .linear)
 
+    // Whether the blur is already built for the current attachment. Repeated
+    // "setupblur" posts are then no-ops: rebuilding a UIVisualEffectView blur
+    // is an expensive filter regeneration on the main thread, and the
+    // notification fans out to EVERY live instance per post (posted per
+    // willDisplay, per menu navigation, per foreground) — during a wallet
+    // reset's academy re-render that multiplied into a full main-thread hang
+    // (watchdog kill), even after the per-instance animator-accumulation fix.
+    // The blur is appearance-independent (overrideUserInterfaceStyle .light),
+    // so the only event that genuinely needs a rebuild is returning from the
+    // background, where iOS can drop a paused animator's effect — handled by
+    // clearing the flag on willEnterForeground.
+    private var isConfigured = false
+
     override func didMoveToSuperview() {
         guard let superview = superview else {
             // Removed from the hierarchy (removeBlur / cell reuse): stop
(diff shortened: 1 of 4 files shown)


RubenWaterman (commented):
Reviewed the PR's own three commits. This rebase was the design-collision one — two real conflicts plus a silent bad composition worse than either — and suite testing on the rebased branch then flushed out two more latent crashes, both fixed on the branch. Details below. The PR is mergeable and scoped to its real files.

## What the PR does well

- **The tap pass-through fix (90eaa48) is the better answer to the swallowed-tap bug.** It supersedes the interim deferred-reveal fix from #40's review: the PIN/signup screen is usable while the intro animation plays (faster unlock for real users), and automation taps can't be swallowed by the cover.
- The animation restructure (4dddc8c) into named spring-damped stages makes the launch sequencing debuggable for the first time, and the end-of-slide dark flash fix (background fade) is real polish. `animationContainer.isHidden = true` at the end is good hygiene — it also removes the container from the accessibility tree.
- The `.light` override for the locked-lesson blur (8cd9132) fixes the too-dark dark-mode blur and — as it turned out — makes a deeper fix possible (see below).

## The reconciliation (6eab384)

`90eaa48` and #40's `revealPinOrSignup` solved the same bug in mutually exclusive ways, and the rebase composed them into something silently broken: this branch's commit only ADDED the pass-through line, assuming the viewDidLoad alpha assignments from its own history were still present — but #40's fix had removed them, and the restructured animation no longer called `revealPinOrSignup`. As merged, **the pin/signup containers never became visible at all**. Resolved in favor of the pass-through approach (this branch's stated intent, and strictly better): alphas restored in `checkWalletAvailability` with the pass-through contract documented, dead `revealPinOrSignup` deleted. Verified with the clearState → tap article → ArticleVC repro (the one that originally exposed the bug) — passes twice against a rebuilt app.

## Two crashes found by suite runs on this branch, fixed here

1. **e7e687a — BlurEffect fan-out hang.** A Sentry AppHang (main thread fully blocked in setFractionComplete during the wallet-reset academy re-render) showed the residual cost after #39's accumulation fix: "setupblur" fans out to every live instance per post, each doing a full blur rebuild. Because this PR's `.light` override makes the blur appearance-independent, instances now build once per attachment and no-op repeat posts — with the one legitimate rebuild trigger (backgrounding can drop a paused animator's effect) preserved via a willEnterForeground flag reset.

2. **f961e49 — checkOnchainAddressesWithBittr crash across wallet reset.** The suite's removal flows reproducibly hit a force-unwrap assertion: the address-check task suspends in its network await, performWalletReset swaps bittrWallet (onchainAddresses = nil), the task resumes into one of six force-unwraps. Fixed with the same bind-once + identity-guard pattern as the listenForEvents race. The guards also close a subtler hazard: post-reset cache write-backs would have resurrected the old wallet's addresses into the NEXT wallet's cache — a fresh wallet handing out its predecessor's receive addresses.

## Nits

- `overrideUserInterfaceStyle = .light` is set per setupBlur call rather than once — idempotent, harmless.
- `checkWalletRemoval` now fires at widenLogoView's completion (~1s in) — fine: the lockout alert overlays the animation container, and earlier resume of an interrupted wipe is a small win.

## Verification

The clearState launch repro passes 2/2 post-reconciliation; both crash fixes build clean. The real test is a full `test_suite.sh` run — its destructive tail (the removal flows) is what exercised both crashes, and with resume checkpointing each rerun costs one flow. A manual dark-mode launch is worth one look (blur brightness + no dark flash at the end of the slide).

PR #42, opened 2026-07-14
Store PIN and Mnemonic in Keychain instead of UserDefaults

tommelters:
- Claude has run a large **security audit**, with the most critical suggestion being **use Keychain**.
- Remove **mnemonic print-statements** from various VCs as not to leak sensitive data.
- **Store PIN and Mnemonic in Keychain**, available on this device only, upon device unlock.
- **Migrate** existing users from UserDefaults to Keychain.
- In checkWalletAvailability(), handle cases in which **Keychain cannot be read**. Allow user to retry.
- Cleanup of **restoreButtonTapped()** function in RestoreVC.
- In **storeMnemonic()**, handle cases in which a new mnemonic cannot be stored. Allow user to retry.

### DIFF
--- a/ios/bittr.xcodeproj/project.pbxproj
+++ b/ios/bittr.xcodeproj/project.pbxproj
@@ -136,6 +136,7 @@
 		05C5A0ED2A33252C004CC671 /* Transfer1ViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05C5A0EC2A33252C004CC671 /* Transfer1ViewController.swift */; };
 		05C5A0EF2A334E30004CC671 /* Transfer3ViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05C5A0EE2A334E30004CC671 /* Transfer3ViewController.swift */; };
 		05CF0FE62D32557300DF2D9F /* GraphView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05CF0FE52D32557300DF2D9F /* GraphView.swift */; };
+		05CF63B93004ECB800E55351 /* SecureStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05CF63B83004ECB800E55351 /* SecureStore.swift */; };
 		05CF933A2A1E711B0084A221 /* Signup3ViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05CF93392A1E711B0084A221 /* Signup3ViewController.swift */; };
 		05D5D0E72A66825000A1E771 /* BitcoinManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05D5D0E62A66825000A1E771 /* BitcoinManager.swift */; settings = {COMPILER_FLAGS = "-fno-objc-arc"; }; };
 		05D5D0EA2A66836300A1E771 /* LightningStorage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05D5D0E92A66836300A1E771 /* LightningStorage.swift */; };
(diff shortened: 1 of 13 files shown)


RubenWaterman (commented):
Reviewed the PR's seven commits in depth — this is the security-critical one, and it holds up well. Conflicts were resolved by rebasing onto android (one conflict: this PR's three-state walletSecretsPresence switch correctly supersedes the two-state launch check while keeping #41's pass-through documentation). Five follow-up commits were added during review, all discussed with and approved by Ruben; details below.

## What this PR gets right — and it gets a lot right

- **The migration design is excellent**: verified-write-then-delete (the UserDefaults copy is removed only after the Keychain copy is confirmed readable), both eager (every launch) and lazy (self-healing inside every read) — so even a partial migration converges safely. This is how you migrate a secret that must never be lost.
- **The three-state presence check is the load-bearing safety insight**: an unreadable Keychain (device locked, transient errors, iOS prewarming) is NOT "no wallet", and the old two-state logic would have answered it with deleteClientInfo + signup. `.unavailable` → non-destructive retry with the protectedDataDidBecomeAvailable observer is exactly right, and the user-facing copy is thoughtful.
- **Fail-loudly on unsaved seeds**: storeMnemonic throws, and both wallet creation and restore abort with an alert rather than proceeding with a seed that only exists in memory — closing a real fund-loss path.
- SecureStore: atomic add-or-update (never delete-then-add), explicit non-synchronizable, kSecUseDataProtectionKeychain, per-environment service isolation. Recovery-phrase prints removed.

## Findings addressed during review (commits on this branch)

1. **Accessibility class vs background modes** → 4f2f817 (+06cc282). The uniform WhenUnlockedThisDeviceOnly contradicted Info.plist's fetch/remote-notification background modes: a locked-device wake reaching getMnemonic() got nil and background payment processing silently degraded — a regression vs UserDefaults for exactly the held-HTLC flows the LSP architecture depends on. Now split per actual usage: **mnemonic + swap keys = afterFirstUnlockThisDeviceOnly** (background node start works; still never backed up / never synced / never on another device), **PIN stays whenUnlockedThisDeviceOnly** (foreground-only read keeps the strictest class free). The trade is recorded in the commit message; stored items are re-stamped to their intended class by the launch migration. (#4's WhenPasscodeSetThisDeviceOnly was considered and rejected: no-passcode users couldn't store a wallet, and iOS deletes such items when a passcode is turned off.)
2. **In-flight swap refund key** (harvested from #4, rebuilt on SecureStore) → 0013009. The ongoingswap UserDefaults dictionary carried the Boltz refund key in plaintext; it now lives in the Keychain (swapkey_<boltzID>, verified write, strip-on-confirmation, inline fallback if the write fails — a plaintext key beats a lost refund key), with eager + lazy migration of existing data and a swapkey_* sweep on wallet reset.
3. **Swap JSON files: encrypted at rest, deliberately plain-text in content** → 153d429 + 1f94381. The files (and the Download button's temp copy) are written with completeFileProtectionUntilFirstUserAuthentication, and existing files are stamped once at launch. The comment now records the policy explicitly: the plain-text refund key is the user-facing emergency artifact for Boltz's rescue flow and must not depend on a working app or Keychain — #4 stripped it from these files, which would have made the emergency file fail exactly in emergencies. Note: the app's Documents folder is not exposed in the Files app (no UIFileSharingEnabled — and it must stay that way while LDK's storage lives in Documents); the in-app Download button is the export route.
4. SecureStore.remove now includes kSecUseDataProtectionKeychain, matching set/get.

## Remaining items — deliberate, for the record

- **Downgrade after migration** (TestFlight rollback): an old build sees empty UserDefaults, concludes "no wallet", and shows signup. The seed survives in the Keychain and re-upgrading recovers it — no loss, but scary UX. Pin the minimum version once this ships.
- **Delete + reinstall semantics change**: Keychain items survive app deletion, so a reinstall finds the seed while LDK's channel state died with the app — a resurrected half-wallet (onchain recoverable, channels not). Suggest a follow-up: detect first-run-after-reinstall (Keychain present, UserDefaults marker absent) and route through an explicit recovery flow instead of silently showing the PIN gate.
- **PIN raw + 10-attempt counter in UserDefaults** — already noted in the code as follow-up hardening. Sharpening: backups restore UserDefaults but not ThisDeviceOnly Keychain items, so a backup-restore cycle resets the fail counter while keeping the PIN — unlimited attempts at a 4-digit PIN for someone with backup access. Top of the hardening list.
- **PR #4 disposition**: its mnemonic/PIN work is superseded by this PR's stronger implementation (atomic writes vs delete-then-add; no unconditional migration flag; no silent permanent UserDefaults fallback; per-environment service). Its one unique contribution — the swap-key protection — is harvested and improved in 0013009/153d429. Recommend closing #4 with credit.

## Verification

Rebased branch builds; clearState+clearKeychain launch → signup repro passes (4/4, re-verified after every commit). Still needed before release and not simulatable here: **a real upgrade test** — install a pre-#42 build, create a funded regtest wallet, upgrade to this build, verify the wallet and PIN survive — plus a full test_suite.sh run (the suite's clearKeychain flows are already compatible). The upgrade test is the one that guards the backwards-compatibility risk this PR carries.

RubenWaterman:
Addendum to the review: the "PIN raw + attempt counter in UserDefaults" item listed under remaining hardening has now been addressed on this branch (60f29c1), at Ruben's request:

- The PIN is stored as a salted SHA-256 record behind verifyPin()/hasPin() — PinViewController no longer holds the raw value at all. Legacy raw values (UserDefaults or earlier builds of this branch) are re-hashed by the launch migration, with a verify-time upgrade as backstop.
- The wipe-after-10 counter moves to the Keychain, closing the backup-restore counter-reset attack. Migration keeps the higher of legacy/current counts so a restored older value can never lower a live one; reads fail open so a transient Keychain error can't wipe a wallet.

Both come with honest-scope comments in the code: hashing doesn't make a 4–8 digit keyspace brute-force resistant (the counter and the mnemonic are the real guards), and a full same-device erase-and-restore cycle still buys 9 attempts per multi-minute round — the change turns a seconds-long plist edit into that, it doesn't eliminate it.

Remaining pre-merge verification is unchanged: the real-device upgrade test (pre-#42 build → funded wallet → upgrade → wallet + PIN survive, now including "PIN still verifies after re-hash") and a full test_suite.sh run.

PR #45, opened 2026-07-21
Various additional guards during wallet loading and resetting

tommelters:
- During wallet removal, only remove **seed and pin details** as the last step.
- Updates to **BittrServiceError** distinctions.
- Improve **getFeeEstimates** to not crash when receiving unexpected data.
- Additional guards against **ldkNode==nil**.
- Guard against **double LDKNode starts**.
- **Thread improvement** for loadWalletData.
- Guard against **double wallet removal**.

### DIFF
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -210,11 +210,16 @@ extension BitcoinManager {
     }
     
     func lightSyncBdkWallet() -> Bool {
+        guard let bdkWallet = self.bdkWallet,
+              let electrumClient = self.electrumClient,
+              let connection = self.connection else {
+            return false
+        }
         
         // Create sync request.
         let syncRequest:SyncRequest
         do {
-            syncRequest = try self.bdkWallet!.startSyncWithRevealedSpks().build()
+            syncRequest = try bdkWallet.startSyncWithRevealedSpks().build()
         } catch {
             self.handleError(error: error, row: 570)
             return false
@@ -223,7 +228,7 @@ extension BitcoinManager {
         // Create update.
         let update:Update
         do {
-            update = try self.electrumClient!.sync(
+            update = try electrumClient.sync(
                 request: syncRequest,
                 batchSize: UInt64(25),
                 fetchPrevTxouts: true
@@ -233,16 +238,21 @@ extension BitcoinManager {
             return false
         }
         
+        // Ensure bdkWallet hasn't been cleared in the meantime.
+        guard self.bdkWallet === bdkWallet else {
+            return false
+        }
+        
         // Apply update to wallet.
         do {
-            try self.bdkWallet!.applyUpdate(update: update)
+            try bdkWallet.applyUpdate(update: update)
         } catch {
             self.handleError(error: error, row: 594)
         }
         
         // Persist update to wallet.
         do {
-            _ = try self.bdkWallet!.persist(connection: self.connection!)
+            _ = try bdkWallet.persist(connection: connection)
         } catch {
             self.handleError(error: error, row: 603)
         }
--- a/ios/bittr/BitcoinManager.swift
+++ b/ios/bittr/BitcoinManager.swift
@@ -482,7 +482,7 @@ class BitcoinManager {
         }
     }
     
-    func getFeeEstimates() async -> NSDictionary? {
+    func getFeeEstimates() async -> FeeEstimates? {
 
         let feeDictionary: NSDictionary
         do {
@@ -516,7 +516,17 @@ class BitcoinManager {
             return nil
         }
 
-        return feeDictionary
+        // A response missing any of the three rates is unusable — treat it as a
+        // failed fetch rather than handing back a half-filled struct that every
+        // call site would have to re-check.
+        guard let fastest = feeDictionary["fastestFee"] as? Double,
+              let hour = feeDictionary["hourFee"] as? Double,
+              let economy = feeDictionary["economyFee"] as? Double else {
+            Log.info("Fee estimate response was missing one or more expected rates.")
+            return nil
+        }
+
+        return FeeEstimates(fastest: fastest, hour: hour, economy: economy)
     }
     
     func syncWallets() throws {
@@ -724,6 +734,14 @@ class BitcoinManager {
         return "m/84'/\(coinType)'/\(account)'/\(change)/\(addressIndex)"
     }
     
+}
+
+/// Recommended on-chain fee rates in sat/vByte, already normalized by
+/// `getFeeEstimates()`: each rate is at least 1 and rounded, as LDKNode requires.
+struct FeeEstimates {
+    let fastest: Double
+    let hour: Double
+    let economy: Double
 }
 
 enum WalletError: Error {
--- a/ios/bittr/BitcoinManager.swift
+++ b/ios/bittr/BitcoinManager.swift
@@ -425,26 +425,24 @@ class BitcoinManager {
     }
     
     func getFeeEstimates() async -> NSDictionary? {
-        
-        var receivedDictionary:NSDictionary
+
+        let feeDictionary: NSDictionary
         do {
-            receivedDictionary = try await withCheckedThrowingContinuation { continuation in
+            feeDictionary = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<NSDictionary, Error>) in
                 Task {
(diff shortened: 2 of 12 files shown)


RubenWaterman (commented):
Reviewed this and pushed five commits to the branch addressing what came out of it. Details below — the findings are recorded as they stood before those commits, with what changed noted under each.

## Scope note

`gh pr diff 45` shows 856/308 across 21 files, but the branch forked from `f2e17d3`, before #42 (Keychain migration) landed on `android`. The net-new diff against current `android` was **196/106 across 12 files** — `SecureStore.swift`, most of `CacheManager.swift`, `RestoreViewController.swift` and `PinViewController.swift` are already merged and were reviewed there. This review covers only the net-new part. Worth rebasing so reviewers see the real diff.

## Overview

Defensive hardening in three groups:

1. **Crash removal** — `ldkNode!`/`bdkWallet!` force-unwraps replaced with `guard let` + `WalletError.walletNotInitiated`; `getFeeEstimates` no longer force-casts API values or resumes a continuation twice; the four fee-estimate call sites key-check instead of `as! Double`.
2. **Wipe ordering** — `performWalletReset` deletes the seed *last*, only after node stop + document deletion succeed, and surfaces a new `removalfailed` alert otherwise.
3. **Single-flight guards** — `isRemovalInFlight` around channel close, `claimNodeStart()`/`endNodeStart()` around LDK node startup.

## What's clearly right

- **`BittrService.swift`** — `throw BittrServiceError.other("Invalid URL" as! Error)` was an unconditional runtime crash (`String` is not `Error`) on every invalid-URL and unsynced-wallet path. Replacing it with real `.invalidURL` / `.unsyncedWallet` cases is the single best fix here. No `as! Error` remains in the target.
- **`getFeeEstimates`** — the old loop could call `continuation.resume(returning:)` inside the loop *and* again after it (fatal double-resume), and mutated an `NSMutableDictionary` while enumerating it. Both gone, and all four callers were updated consistently — none were missed.
- **`lightSyncBdkWallet`** (`BDKManager.swift:213`) — hoisting the three handles and re-checking `self.bdkWallet === bdkWallet` after the blocking network sync is exactly the right shape for "the wipe may have niled this mid-flight".
- **Seed-last ordering** in `performWalletReset` — correct, and the failure path leaving the seed intact is the safe direction.

## Findings

### High — `.startInFlight` was a silent dead end, and the watchdog's "Try again" always hit it

`StartLightning.swift` returned from that branch with no UI change and no completion path, and `isStartingNode` was only cleared after `startLightning()` returned.

The watchdog exists precisely for the case where `startLightning()` *hangs*:

1. `startWallet()` → `.proceed`, `isStartingNode = true`, node build hangs.
2. 15s later the watchdog shows the retry alert. `isStartingNode` is still `true`.
3. User taps Try again → `restartLightning()` → `hideAlert()` → `startWallet()` → `.startInFlight` → returns silently.

The alert was now dismissed, no new watchdog was armed, and nothing would ever re-arm one — recovery required a force-quit. The flag is process-global, so for as long as the start stayed stuck, `startWallet()` from `PinViewController:213`, `ReloadWallet:24`, `Signup6:57` and `RestoreViewController:314` were all silent no-ops too. Same shape hit the lockout wipe: `startWalletInBackground` raises `fullViewCover` to 0.8 and starts the spinner *before* calling `startWallet()`, so a `.startInFlight` return left a permanent full-screen cover.

**Fixed** in 3aad878. `claimNodeStart` now takes an `onInFlightStart` completion that is attached to the start already running and fired with its outcome, and the in-flight branch arms its own watchdog so a hang still surfaces a retry.

One consequence worth flagging for re-review: since several callers can now be attached to a single start, they each get a success callback, where previously the duplicates were dropped. `continueStartWallet` is global work (`syncWallets`, `loadWalletData`, `startBDK`), not per-caller, so the same commit guards it against running twice and releases that guard via `defer` when the post-start block exits.

### High — the successful coop-close path never released `isRemovalInFlight`

`ResetApp.swift` showed the "still closing / Try again" alert after a successful close broadcast but left `isRemovalInFlight == true`; it was only cleared on the failure and fully-swept paths.

So on the lockout path: close broadcast succeeds → user taps Try again → `startWalletInBackground` (cover + spinner up) → `restoreWalletTapped` → if `getActiveChannel()` still returns the channel (`QuestionViewController.swift:131` filters on `isChannelReady`, which stays true during shutdown negotiation, before the closure event fires), the guard returned silently. Cover and spinner stayed up, no alert, and nothing cleared the flag for the rest of the session.

**Fixed** in f7b3617 — the guard is released where the terminal alert goes up. User action is required at that point, so a fresh attempt is legitimate.

### Medium — the wipe-failure path left the app half-reset

`performWalletReset` cleared `resettingPin`, `removingWalletForIncorrectPin`, `walletSync` and `bittrWallet` *before* the async cleanup, and the `cleanupSucceeded == false` branch restored none of it — leaving a zeroed in-memory wallet and both removal flags cleared, outside the PIN reset and lockout flows alike, while the seed and every file it described still existed. Survivable, since the seed and the resume flag persist, but the in-session state described a wipe that never happened.

**Fixed** in b5e57f8 — the flags are remembered and restored on failure, and `bittrWallet` is only cleared once the wallet is actually gone. The sync timer still stops first, deliberately: a sync running against a half-deleted wallet is worse than losing the timer, and the failure alert tells the user to reopen the app.

On the related question of whether skipping `setWalletRemovalInProgress(true)` for the `removingWalletForIncorrectPin` path leaves the lockout wipe unresumable — **it doesn't, no change needed.** `CoreViewController.swift:220` re-arms it at launch off `getFailedPinAttempts() >= 10`, independently of the removal-in-progress flag. The comment about the seed resuming on next launch is accurate for both paths.

### Low

- **`LoadWalletData.swift`** — `guard let node = ... else { return }` sat after `CacheManager.storeTxoID` had already run, so a nil node left the cache holding a funding txo ID from a snapshot that was never applied. **Fixed** in b7c74ca.
- **`getFeeEstimates` returned `NSDictionary?`** — all four call sites repeated the same `["fastestFee"] as? Double` dance over rates the function had already validated, and the return had a third state nobody handled: non-nil but missing keys. **Fixed** in f550d38 with a `FeeEstimates` struct. Note the behaviour change: a response missing *any* of the three rates now reports as a failed fetch, where the three `fastestFee`-only callers would previously have proceeded on a partial response.
- **claim/release were split across files** — any future path calling `startLightning()` directly would bypass the guard, and any early return between the two would leak the flag permanently. **Fixed** in 3aad878; both sides now live in `runClaimedNodeStart`, which releases via `defer`.
- **`claimNodeStart` calls `status()` under `nodeStartLock`** — an FFI call into the LDK node while holding a mutex. **Left as-is deliberately.** It's safe today (nothing else takes this lock) and every way of moving the call outside the lock opens a real double-start window, which is a worse trade than the habit is worth. Flagging it so it's a known, deliberate choice rather than an oversight.
- **`LoadWalletData` main-thread hop** — the `Thread.isMainThread ? apply() : async` branch is dead code today; all three callers (`StartLightning`, `BitcoinManager:445`, `HandlePaymentNotification:573`) are already on main. Harmless, but note the else branch makes `bittrWallet` writes asynchronous, so a future off-main caller reading it right after would see stale values.

## Style / conventions

Consistent with the codebase — `guard let` + early return, `Log.info` breadcrumbs, Sentry captures with `context` extras. No concerns.

## Testing

No tests here, and `bittrTests` is still the stock Xcode template (`testExample` + `testPerformanceExample`, nothing real), so adding any would mean this PR establishes the repo's first test suite — deliberately left as a follow-up rather than riding along on a bugfix PR. The two pieces worth covering when that happens are the `claimNodeStart` → `runClaimedNodeStart` state machine (double-claim, attached completions, hang) and `getFeeEstimates` against malformed payloads. Both are exactly the logic this PR is meant to harden.

Given the funds-loss stakes, the wipe paths still want a manual matrix before merge, and the fixes above have **only been verified to compile — none of them have been exercised at runtime**:

- open channel + lockout
- close-broadcast, then Try again
- node-start hang, then Try again (the `.startInFlight` path)
- two racing `startWallet` callers (the new `continueStartWallet` guard)
- cleanup failure

## Verdict

The crash fixes are unambiguously good. The two single-flight guards were the risky part — both had a state where the guard was set and never cleared, leaving the user behind a full-screen cover with no action available, on the lockout path where they already can't get into their wallet. Those are fixed on the branch now.

Remaining before merge: rebase onto current `android` so the diff reflects the real 196/106 scope, and run the manual matrix above.

PR #46, opened 2026-07-21
Allow clearing a transaction note to delete it

tommelters:
In **TransactionVC**, emptying the **note text field** and tapping Save now **deletes the stored note** and reverts the card to the "Add a note" button, instead of silently keeping the old note.

### DIFF
--- a/ios/bittr/CacheManager.swift
+++ b/ios/bittr/CacheManager.swift
@@ -808,6 +808,21 @@ class CacheManager: NSObject {
         }
     }
     
+    static func deleteTransactionNote(txid:String) {
+
+        let envKey = EnvironmentConfig.cacheKey(for: "transactionnotes")
+
+        let defaults = UserDefaults.standard
+        let cachedTransactionNotes = defaults.value(forKey: envKey) as? NSDictionary
+        if let actualCachedNotes = cachedTransactionNotes {
+            // Notes have been cached.
+            if let actualMutableNotes = actualCachedNotes.mutableCopy() as? NSMutableDictionary {
+                actualMutableNotes.removeObject(forKey: txid)
+                defaults.set(actualMutableNotes, forKey: envKey)
+            }
+        }
+    }
+
     static func getTransactionNote(txid:String) -> String {
         
         let envKey = EnvironmentConfig.cacheKey(for: "transactionnotes")
--- a/ios/bittr/Transaction/TransactionViewController.swift
+++ b/ios/bittr/Transaction/TransactionViewController.swift
@@ -640,8 +640,13 @@ class TransactionViewController: UIViewController {
         NSLayoutConstraint.deactivate([self.noteStackHeight])
         self.noteStackHeight = NSLayoutConstraint(item: self.noteStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 0)
         NSLayoutConstraint.activate([self.noteStackHeight])
-        self.addANoteStack.alpha = 1
-        self.addANoteStackHeight.constant = 45
+        // Keep "Add a note" hidden on the Bittr payout summary, which suppresses
+        // it deliberately — reverting to it there would put the button on a
+        // screen that never offers one.
+        if !self.showConfetti {
+            self.addANoteStack.alpha = 1
+            self.addANoteStackHeight.constant = 45
+        }
         self.view.layoutIfNeeded()
     }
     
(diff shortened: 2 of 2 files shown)


RubenWaterman (commented):
Reviewed this and pushed one commit (`e8e1367`) with the two small things that came out of it. Findings below are recorded as they stood against `528bee6`, with what changed noted under each.

## Overview

Emptying the note field in `TransactionViewController` and tapping Save now deletes the stored note and reverts the card to the "Add a note" button, instead of silently keeping the old one. Two additions:

- `CacheManager.deleteTransactionNote(txid:)` — removes the key from the `transactionnotes` dictionary in `UserDefaults`.
- `TransactionViewController.hideNoteStack()` — the mirror of the existing `showNoteStack()`, plus an `else` branch in `noteButtonTapped` that calls both.

Purely additive, no behaviour change to any existing path.

## The risk worth checking, and it's clean

The obvious way for this change to go wrong is if the alert's completion fires on **Cancel** — a user opening a note, tapping Cancel, and silently losing it. Traced it, and it's fine:

- `AlertManager.swift:531` — `onSave` is wired only to the `index == 1` button. Cancel (`index == 0`) calls `dismiss()` and nothing else.
- `textField.returnKeyType = .done` (`:441`), but there's no `UITextFieldDelegate`, no `shouldReturn`, and no action on the field. Done only lowers the keyboard — the comment in `dismiss()` at `:475` confirms that's the expected path.
- No tap-outside-to-dismiss gesture.

So Save is the only route into the new branch.

The guard is also correctly `trimmingCharacters(in: .whitespacesAndNewlines) != ""`, so a whitespace-only note deletes rather than storing a blank one that would render as an empty note card.

## What's right

- **`deleteTransactionNote` is an exact structural mirror of `storeTransactionNote`** — same `EnvironmentConfig.cacheKey(for: "transactionnotes")`, same `mutableCopy()` → mutate → `defaults.set` pattern, same nil-guard shape. It reads like it was always there.
- **`hideNoteStack` correctly inverts `showNoteStack`**: `.equal` constant 0 to collapse, versus `.greaterThanOrEqual` 0 to let content drive the height. The deactivate-then-reassign dance on the `weak` outlet is the pre-existing idiom, and the new constraint is retained by the view hierarchy on `activate`, so the weak reference stays alive.
- **`hideNoteStack` is actually more thorough than the setup path it restores to.** The "no note" branch at `:619` only touches `addANoteStack` and leaves `noteStack` at its storyboard default; `hideNoteStack` explicitly collapses both.

## Findings

### Low — `hideNoteStack` unconditionally revealed the one thing the confetti path suppresses

`hideNoteStack()` set `addANoteStack.alpha = 1` / height 45 with no `showConfetti` check, while setup at `:611` deliberately hides that stack for the Bittr payout summary.

Worth being precise: **this was not reachable.** `noteStack` defaults to `alpha="0.0"` with a zero-constant height (`Main.storyboard:7697`, constraint `Jh7-UN-psp`), and the `showConfetti` branch never calls `showNoteStack()` — so on that screen both stacks are collapsed and there is no note button to tap.

But it was unreachable by coincidence of a storyboard default rather than by intent, and `SendLightning.swift:382` already stores notes automatically at payment time, so "show the existing note on the payout summary" is a plausible future change. That change would quietly resurrect a button the confetti path is meant to suppress.

**Guarded** in `e8e1367`, so the invariant sits next to the code that would break it. Defensive, not a live bug fix.

### Nit — the guard trimmed, but the store didn't

```swift
if noteText.trimmingCharacters(in: .whitespacesAndNewlines) != "" {
    CacheManager.storeTransactionNote(txid: ..., note: noteText)   // untrimmed
```

`"  hi  "` persisted with its padding and redisplayed padded on every later visit. Pre-existing, but this PR owns that branch now.

**Fixed** in `e8e1367` — hoisted into a `trimmedNote` local so the value checked is the value stored and displayed. Two caveats: it isn't retroactive, so notes already saved with padding stay padded until edited; and `SendLightning.swift:382` has the same asymmetry on the auto-stored path, left alone as outside this PR's scope.

### Nit — `Transaction.note` is written and never read

`Transaction.swift:20` declares it, `:120` populates it from `getTransactionNote`, and nothing anywhere consumes it. Harmless today, but it's now a stale-data trap: notes can be deleted, and the model field is captured at list-build time, so whoever first renders `transaction.note` in a cell gets a note that may already be gone. Either drop the field or note that `TransactionViewController` reads through `CacheManager` deliberately. **Left alone** — out of scope for this PR.

### Style — the two stack functions are ~90% duplicated

`showNoteStack` and `hideNoteStack` differ in three values (alpha, constraint relation, the `addANoteStack` pair). A single `setNoteStack(visible:)` would halve them and make the two states impossible to drift apart. **Left alone** — reasonable as a follow-up, and #49–#54 suggest that's the house pattern anyway.

## Testing

No tests, and both targets are still stock Xcode templates (`bittrTests.swift` is `testExample` + `testPerformanceExample`; nothing in `bittrUITests` references notes). Same situation as #45 — starting a suite shouldn't be this PR's job.

Worth noting this change is unusually cheap to UI-test whenever that does start: the alert already exposes `alert.button.1` and `alert.textField`, and `labelNote` carries `TestID.Transaction.labelNote`. Clear-and-save is three interactions and one assertion.

The branch **compiles, but none of this has been exercised at runtime**, including my two fixes. Manual matrix before merge:

- add note → reopen → clear → Save → card reverts, and the note is still gone after leaving and re-entering the screen
- clear → **Cancel** → note survives (the code is correct, but it's the one that would hurt)
- whitespace-only note → Save → deletes
- Save on a transaction that never had a note (exercises `hideNoteStack` against the storyboard constraint rather than one `showNoteStack` installed)
- a Bittr payout summary with confetti → "Add a note" stays hidden throughout

## Security

Notes stay plaintext in `UserDefaults` — unchanged here, and this is user-authored metadata rather than a secret, so it sits outside the #42 Keychain migration's scope. The delete path is a marginal privacy improvement. Removing a key isn't a secure erase (the old value can persist in device backups and in the on-disk plist until rewritten), but that's not worth acting on here.

## Verdict

**Approve.** Small, correct, and close enough to the existing patterns that it reads as native to the file. The one thing that could have made it dangerous — Cancel routing into the delete branch — doesn't happen. Both items I'd have asked for are now on the branch; what's left is the manual pass above.

PR #47, opened 2026-07-21
Fix BDK sync scan latch and false-success in light sync

tommelters:
didSyncBdkWallet: the 180s watchdog cleared bdkWalletIsScanning while the un-cancellable fullScan kept running, so a second scan could start against the same non-thread-safe wallet/SQLite and corrupt it. Clear the latch only when the real scan closure finishes; the watchdog now only unblocks the caller's completion.

lightSyncBdkWallet: a failed applyUpdate fell through to `return true`, reporting a failed sync as success and feeding stale coins into swap fee/coin selection. Return false in the catch, as didSyncBdkWallet does.

### DIFF
--- a/ios/bittr/AlertManager.swift
+++ b/ios/bittr/AlertManager.swift
@@ -44,17 +44,22 @@ protocol OnchainSyncFailureReporting: UIViewController {
 
 extension OnchainSyncFailureReporting {
     /// Stops the on-chain sync spinner and tells the user the on-chain sync
-    /// failed and to retry later. Safe to call from any thread — BDK completions
-    /// fire off the main thread.
+    /// failed. Safe to call from any thread — BDK completions fire off the main
+    /// thread.
+    ///
+    /// A scan that timed out gets different copy: the scan is still running and
+    /// can't be stopped, so "try again later" would send the user back to a
+    /// screen that stays blocked no matter how long they wait.
     func presentOnchainSyncFailedAlert() {
         DispatchQueue.main.async {
             self.bdkSpinner.stopAnimating()
+            let timedOut = BitcoinManager.shared.bdkFullScanTimedOut
             // Replace the "syncing" alert (if still visible) with the failure alert.
             self.hideAlert()
             self.showAlert(
                 presentingController: self,
                 title: Language.getWord(withID: "onchainsyncfailedtitle"),
-                message: Language.getWord(withID: "onchainsyncfailed"),
+                message: Language.getWord(withID: timedOut ? "onchainsynctimedout" : "onchainsyncfailed"),
                 buttons: [Language.getWord(withID: "okay")],
                 actions: nil
             )
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -86,7 +86,14 @@ extension BitcoinManager {
     func didSyncBdkWallet(completion originalCompletion: @escaping (Bool) -> Void) {
         Log.info("Will sync BDK wallet.")
         self.bdkWalletIsScanning = true
-        
+        self.bdkFullScanTimedOut = false
+
+        // The watchdog below hands the caller a failure, but it cannot stop the
+        // scan — so `bdkWalletIsScanning` stays set and every scan-gated screen
+        // stays blocked until the app is restarted. Callers surface that via
+        // `bdkFullScanTimedOut` rather than the generic "try again later" copy,
+        // which would be a lie: nothing the user does in this session will help.
+        //
         // NOTE: `timeout` must stay comfortably above a legitimately slow scan
         // (the fullScan below is documented as "seconds to minutes"), or slow
         // syncs get reported as failures. Tune before shipping.
@@ -101,6 +108,7 @@ extension BitcoinManager {
         }
         DispatchQueue.global(qos: .background).asyncAfter(deadline: .now() + timeout) {
             Log.info("BDK full scan timed out after \(Int(timeout))s; failing so callers don't hang.")
+            self.bdkFullScanTimedOut = true
             completion(false)
         }
 
@@ -201,6 +209,17 @@ extension BitcoinManager {
     }
     
     func lightSyncBdkWallet() -> Bool {
+        // Stand down while a full scan is running. Both work through the same
+        // wallet, electrum client and SQLite connection, which serialize
+        // internally — so overlapping them doesn't corrupt anything, it starves
+        // the scan behind a light sync that re-fires every 30s until the scan's
+        // watchdog gives up. BackgroundSync has no idea a scan is in flight, so
+        // the check has to live here.
+        guard !self.bdkWalletIsScanning else {
+            Log.info("Skipping light sync — a BDK full scan is in progress.")
+            return false
+        }
+
         guard let bdkWallet = self.bdkWallet,
               let electrumClient = self.electrumClient,
               let connection = self.connection else {
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -86,15 +86,7 @@ extension BitcoinManager {
     func didSyncBdkWallet(completion originalCompletion: @escaping (Bool) -> Void) {
         Log.info("Will sync BDK wallet.")
         self.bdkWalletIsScanning = true
-
-        // Fire the caller's completion exactly once and clear the scanning flag.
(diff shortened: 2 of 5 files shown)


RubenWaterman (commented):
Reviewed this, rebased it onto current `android`, and pushed a second commit after testing turned up something the original review missed. Findings below are recorded as they stood against `e13bc59`, with what changed noted under each.

## Overview

Two independent fixes in `BDKManager.swift`:

1. **`didSyncBdkWallet`** — the 180s watchdog cleared `bdkWalletIsScanning` while the un-cancellable `fullScan` kept running, so a second scan could start against the same non-thread-safe wallet and SQLite connection. The latch is now cleared only by the real scan closure; the watchdog only unblocks the caller.
2. **`lightSyncBdkWallet`** — a failed `applyUpdate` fell through to `return true`, reporting a failed sync as success and feeding stale UTXOs into swap fee estimation and coin selection. Now returns `false`.

Both diagnoses are correct and both fixes are right. A third fix was added during review — see below.

## Branch state — rebased

As submitted, this branch forked at `f2e17d3` and carried 20 commits already on `android`: #42's Keychain work, and a **pre-rebase copy of #45**. Since #45 was squash-merged as `0f5c654`, `android` had those files *including* their five follow-up fixes; this branch had the versions without them (`runClaimedNodeStart`, `isContinuingStartWallet`, `struct FeeEstimates`, `wasRemovingWalletForIncorrectPin`, `onInFlightStart` were all absent).

All 10 merge conflicts were in exactly those files. Resolving them by taking "ours" would have silently reverted five fixes that landed an hour earlier — a plausible mistake across 10 conflicts in files this PR has no interest in.

Rebased with `git rebase --onto origin/android 7682a6c`. `BDKManager.swift` at the old parent was byte-identical to `android`'s, so it replayed with zero conflicts.

```
before:  21 commits, +858/-318 across 21 files, CONFLICTING
after:    2 commits,  +37/-11   across  5 files, CLEAN
```

## Correctness

**Fix 1 — the latch.** Every exit in the scan closure that returns early clears the flag before calling `completion(false)`: electrum init, the `guard let` snapshot, `startFullScan().build()`, `fullScan`, the `self.bdkWallet === bdkWallet` identity check, `applyUpdate`, and the success path. The `persist` failure deliberately falls through and is cleared at the end. No path leaks the latch set. The change also *reduces* the number of writing threads — the removed line was the only write from main.

Worth noting for anyone reading the diff: on a **successful** scan this change is a no-op. The background closure already sets the flag false immediately before calling `completion(true)`, so the removed line was redundant on every path except the watchdog.

**Fix 2 — the false success.** Correct, and it matches `didSyncBdkWallet`'s shape at the equivalent site.

## The finding the original review missed

I filed the unsynchronized `bdkWalletIsScanning` as a "Low" nit and framed the risk as scan-vs-scan. Testing showed that framing was wrong, and it matters:

```
Will light sync wallet.   x5
BDK full scan timed out after 180s; failing so callers don't hang.
Will light sync wallet.   x6   <- scan still parked
```

`BackgroundSync` fires every 30s (`BackgroundSync.swift:17`) and calls `lightSyncBdkWallet`, which checked only that the handles were non-nil — it **never consulted `bdkWalletIsScanning`**. Both paths work through the same `bdkWallet`, `electrumClient` and `connection`. Those serialize internally, so the two don't corrupt each other; they starve the full scan behind a light sync that re-fires faster than the scan can make progress, until the watchdog gives up. 180s ÷ 30s = 6 light syncs per window, which is exactly the observed cadence — and they continue *after* the timeout, proving the scan is still parked.

So **scan-vs-scan was never the collision happening in practice.** The one that hangs the app is light-sync-vs-scan, every 30 seconds. This PR as originally submitted guarded the former and left the latter untouched.

There was also an interaction this PR *worsened*: because the watchdog no longer clears the latch, after a timeout `bdkWalletIsScanning` stays set forever — but the timer kept light-syncing anyway, since it didn't check the latch. Permanent zombie scan, permanent 30s contention loop, permanently blocked Send and Swap.

**Fixed** in `e2743ca`. The guard lives in `lightSyncBdkWallet` rather than `BackgroundSync` because `lightSync` also does LDK-side balance and payment checks that are fine during a BDK scan — only the BDK call needs to stand down. It has a single caller, so that's the whole surface.

Confirmed working: the hang no longer reproduces.

## The timeout is now visible

A timed-out scan leaves the session permanently scan-blocked, and the existing copy said *"Please try again later"* — untrue, since nothing the user does in that session helps. The launch path (`StartLightning.swift`) said nothing at all, so a user on the home screen just found Send and Swap quietly broken.

`e2743ca` adds `bdkFullScanTimedOut`, switches `presentOnchainSyncFailedAlert` to accurate copy when it's set, and gives the previously-silent launch path its own alert.

Deliberately **not** shown from the watchdog itself: `completion(false)` propagates to callers that already alert, and a second alert queued from the watchdog would race theirs, with main-queue ordering handing the win to the caller anyway. Driving the message off a flag means each site alerts exactly once with no ordering dependency.

## Remaining findings

### Low — `bdkWalletIsScanning` is still an unsynchronized `Bool` across queues

Written from a background queue in the scan closure, read from main in `SendViewController` and `SendableAmount`, read from a background queue in `startBDK` and now in `lightSyncBdkWallet`. No synchronization.

Pre-existing, but this latch is now genuinely load-bearing — it's the only thing keeping two BDK consumers off the same wallet. In practice the reads that matter happen after a `DispatchQueue.main.async` hop, which is a barrier, so this isn't the current bug. Given `BitcoinManager` gained `nodeStartLock` for the same class of problem in #45, the same treatment would be consistent. `bdkFullScanTimedOut` inherits the same caveat.

### Low — `SendViewController.swift:238` registers nothing while waiting

```swift
} else {
    Log.info("Waiting for BDK wallet to finish scanning.")
}
```

Opening Send during a scan puts up the spinner and "you can send 0 satoshis", then waits on exactly one external caller — `StartLightning.swift:206`, `self.homeVC?.sendVC?.setSendAllLabel()`. Two optional chains; if either is nil at that moment the screen stays stuck with no recovery.

This is what made the original hang look like a lost callback rather than a stalled scan. Much less likely to bite now that scans complete, so **left alone** — but it's a real dead end, and the natural fix is the shape used for `.startInFlight` in #45: attach to the in-flight scan rather than hoping to be called back.

Related: `HomeToSend` sets `sendVC.coreVC` and `self.sendVC`, but unlike `HomeToMove` immediately above it (`moveVC.homeVC = self`) it never sets `sendVC.homeVC`. Doesn't break this path, but the wiring is incomplete.

### Nit — "Tune before shipping" is still in the code

`// NOTE: timeout must stay comfortably above a legitimately slow scan … Tune before shipping.` Inherited, not introduced here, but 180s is now the value deciding how long before a user is told to restart the app. Worth confirming or dropping the TODO.

### Nit — the rationale comment was deleted rather than updated

The first commit removes eight lines explaining why the watchdog exists and what it races against, replacing them with nothing (and a trailing-whitespace line). That comment is *more* necessary now, since the watchdog's contract changed: it no longer clears the latch, and the reason — the scan can't be cancelled — is exactly what stops someone "fixing" it back. `e2743ca` puts an explanation back above the timeout, but the original wording was better on the race itself.

### Observation — `persist` failure still reports success

In both functions a failed `persist` is logged and then reported as a successful sync. Consistent between the two, and defensible since the in-memory wallet *is* updated — but the wallet silently diverges from disk until the next successful persist. Not this PR's problem; noted since it's the last path with the shape the PR fixes elsewhere.

## Testing

No automated tests; both targets remain stock Xcode templates, same as #45 and #46. Neither fix is cheap to unit-test (both need a live electrum).

Manually exercised: the light-sync starvation hang reproduced before `e2743ca` and does not after — a cold start now logs `Skipping light sync — a BDK full scan is in progress.` and the scan completes.

**Not** exercised: the timeout path itself (needs a genuinely unreachable electrum), so the new alert copy and the `bdkFullScanTimedOut` plumbing are compile-verified only. Worth pointing `EnvironmentConfig.electrumURL` at a blackhole once before merge to confirm the alert appears on both the launch and Send paths.

## Verdict

**Approve.** The two original fixes are correct and well-diagnosed. The one that actually resolves the reported hang is the third, added during review — and it's worth recording that the PR as submitted guarded the collision that doesn't happen while leaving the 30-second one untouched, because that's an easy conclusion to draw the wrong way round from the commit message alone.

PR #48, opened 2026-07-21
Guard ConfirmSendVC confirm button against double-tapping

tommelters:
In the **ConfirmSendVC**, confirmButtonTapped started the send spinner without blocking further taps, so a rapid double-tap fired two sendPayment/broadcast calls. 

LDK dedups BOLT11 by payment hash, but BOLT12/zero-amount and on-chain sends are timing-dependent. 

Return early while confirmSpinner is animating, mirroring SwapViewController.nextTapped; the spinner is cleared on every success/failure/retry path, so it doubles as the in-flight latch.

### DIFF
--- a/ios/bittr/Move, Send, Receive/SendVC/Confirm/ConfirmSendViewController.swift
+++ b/ios/bittr/Move, Send, Receive/SendVC/Confirm/ConfirmSendViewController.swift
@@ -282,6 +282,7 @@ class ConfirmSendViewController: UIViewController {
     }
     
     @IBAction func confirmButtonTapped(_ sender: UIButton) {
+        if self.confirmSpinner.isAnimating { return }
         guard self.checkInternetConnection() else { return }
         
         if self.sendVC!.onchainOrLightning == .onchain {
--- a/ios/bittr/Move, Send, Receive/SendVC/SendOnchain.swift
+++ b/ios/bittr/Move, Send, Receive/SendVC/SendOnchain.swift
@@ -178,6 +178,14 @@ extension ConfirmSendViewController {
     }
     
     @objc func performOnchainTransaction() {
+        // An on-chain send reaches the broadcast from the confirmation alert's
+        // Confirm button, not from confirmButtonTapped — which shows that alert
+        // and returns without ever starting the spinner. So the guard on the
+        // button can't catch this; it has to sit here, where the spinner goes up.
+        // sendOnchainPayment below blocks main, so a second tap queues behind it
+        // and gets delivered once the broadcast is already under way.
+        if self.confirmSpinner.isAnimating { return }
+
         self.hideAlert()
         
         // Start spinner.
--- a/ios/bittr/Notifications/HandlePaymentNotification.swift
+++ b/ios/bittr/Notifications/HandlePaymentNotification.swift
@@ -395,9 +395,17 @@ extension CoreViewController {
                     let receiveVC = (self.homeVC!.presentedViewController as? ReceiveViewController ?? self.homeVC!.moveVC?.presentedViewController as? ReceiveViewController)
                     let swapVC = self.swapVC
                     
-                    // Update views.
+                    // Update views. The confirm spinner has to be stopped here
+                    // too, the way addNewPaymentToTable does on success: a
+                    // payment LDK accepts and then fails asynchronously
+                    // (routeNotFound, retriesExhausted, recipientRejected)
+                    // leaves the user on the confirm screen, and the spinner
+                    // now gates the Confirm button — so leaving it running
+                    // would strand them with no way to retry.
                     sendVC?.nextLabel.alpha = 1
                     sendVC?.nextSpinner.stopAnimating()
+                    sendVC?.confirmSendVC?.confirmLabel.alpha = 1
+                    sendVC?.confirmSendVC?.confirmSpinner.stopAnimating()
                     sendVC?.resetFields()
                     
                     // Parse failure reason.


RubenWaterman (commented):
Reviewed this and pushed a second commit (`dd42fb2`) addressing what came out of it. Findings are recorded as they stood against `ca48efa`, with what changed noted under each.

## Overview

Adds one line to `ConfirmSendViewController.confirmButtonTapped`:

```swift
if self.confirmSpinner.isAnimating { return }
```

The intent is to stop a rapid double-tap from firing two sends, reusing the existing spinner as the in-flight latch.

## What's right

- **The style matches the cited precedent exactly.** `SwapViewController.swift:133` is `if self.nextSpinner.isAnimating { return }`. Checked, not assumed.
- **It is correct for the lightning path.** `confirmButtonTapped` calls `performLightningPayment()`, which sets `confirmSendVC?.confirmSpinner.startAnimating()` at `SendLightning.swift:244` **synchronously**, before the `Task {`. A second tap arrives in a later runloop turn and sees `isAnimating == true`. That closes the BOLT12 / zero-amount double-send the description calls out.

## Findings

### High — the guard didn't reach on-chain sends, which the description claims to cover

The description says BOLT12 and *"on-chain sends are timing-dependent"*, implying both. On-chain wasn't covered at all.

`confirmButtonTapped` doesn't broadcast anything on-chain. It calls `confirmSendOnchain()` → `proceedWithOnchainConfirmation()`, which puts up the **"Send transaction?" confirmation alert** and returns — never starting the spinner. The broadcast happens in `performOnchainTransaction`, wired to that alert's Confirm button, and that's where the spinner goes up (`SendOnchain.swift:185`) and where `sendOnchainPayment` runs (`:198`).

So on a double-tap of the Confirm *button*, `confirmSpinner.isAnimating` was still `false` and the guard never fired — you'd get two stacked confirmation alerts. The real double-broadcast risk is double-tapping the **alert's** Confirm: `sendOnchainPayment` runs synchronously on main, so main is blocked through the broadcast and a queued second tap is delivered afterwards, hit-testing against an alert `hideAlert()` has only begun dismissing. Timing-dependent, exactly as the description says — and unguarded.

**Fixed** in `dd42fb2` by guarding `performOnchainTransaction`, where the spinner actually starts.

**Deliberately not applied to `performLightningPayment`.** My first pass suggested it; that would have been a bug. `performLightningPayment` re-enters *itself* at `SendLightning.swift:254` after a successful peer reconnect, while the spinner it started at `:244` is still running — an `isAnimating` guard there would silently drop the retry, leaving the payment unsent, the spinner running and no alert. The alert-driven entry points (`SendLightning.swift:265`, `SwapViewController.swift:288`) are single-shot buttons, and the initial tap is already covered by the guard in `confirmButtonTapped`, so the lightning side needs nothing further.

### High — using the spinner as the latch turned one failure path into a dead Confirm button

The description states the spinner *"is cleared on every success/failure/retry path, so it doubles as the in-flight latch."* That held for three paths out of four:

| Path | Stopped `confirmSpinner`? |
|---|---|
| Success → `.paymentSuccessful` → `addNewPaymentToTable` (`SendLightning:376`) | yes |
| Synchronous throw from `sendPayment` (`SendLightning:323`) | yes |
| Peer reconnect failure (`SendLightning:264`) | yes |
| **Async `.paymentFailed` event** (`HandlePaymentNotification:390`) | **no** |

The `.paymentFailed` handler stopped `sendVC?.nextSpinner`, called `resetFields()` — which only clears two text fields (`SendSwitch.swift:12`) — and showed the failure alert. It never touched `sendVC?.confirmSendVC?.confirmSpinner`, and unlike the success path it doesn't `slideFromConfirmToSend()`, so the user stays on the confirm screen.

That's the *common* failure mode: `routeNotFound`, `retriesExhausted`, `recipientRejected` — payments LDK accepts and then fails asynchronously. Before this PR the spinner would spin but Confirm still worked, so the user could retry. With the guard in place and nothing stopping the spinner, Confirm was dead for the life of that `ConfirmSendViewController` — trading a possible double-send for "cannot retry a failed payment", which is the wrong direction.

**Fixed** in `dd42fb2` — the handler now stops `confirmSpinner` and restores `confirmLabel`, as `addNewPaymentToTable` already does on success.

One judgement call: the spinner is stopped but the view does **not** slide back to the send screen, unlike the success path. Leaving the user on the confirm screen with a working button seems right for a retry, but it's a deliberate asymmetry rather than an oversight — easy to change if the other behaviour is preferred.

### Low — a view property as the concurrency latch

`isAnimating` is UI state, so the latch is only as sound as every completion path remembering to stop it — and the finding above is that failure mode already having happened, on the first path anyone checked. A dedicated `isSending` flag on `ConfirmSendVC` cleared in a single teardown wouldn't depend on that discipline.

Consistency with the existing `nextSpinner` pattern argues for keeping it as-is, so not worth blocking on — but the precedent being copied carries the same latent weakness, and it has now bitten once.

### Nit

`if … { return }` matches `nextTapped:133`, though the following line in the same function uses `guard … else { return }`. Matching the cited precedent is the more defensible choice.

## Branch state — rebased

Originally forked at `f2e17d3`, so the merge-base predated #42, #45, #46 and #47. GitHub still reported it clean, since a single isolated line can't conflict, and my first read was that no rebase was needed.

That changed once the fix above meant editing `SendOnchain.swift`, which *has* diverged on `android` (#45's `FeeEstimates` refactor). Building against the old base would have verified a combination that never ships, so the branch was rebased onto current `android` first — one commit, no conflicts.

## Testing

No automated tests; both targets remain stock Xcode templates, same as #45–#47.

Everything here is **compile-verified only — none of it has been exercised at runtime.** Two cases are worth doing by hand before merge:

- **Double-tap the on-chain confirmation alert's Confirm button.** This is the path that can broadcast twice, i.e. actually send funds twice. It was unguarded before `dd42fb2`.
- **A lightning payment that fails asynchronously** (no route available, or one that reaches `retriesExhausted`). Confirm the button works again after the failure alert — this is the regression the original one-liner would otherwise have shipped.

Also worth a pass: rapid double-tap Confirm on lightning (one payment), and on on-chain (one alert, not two stacked).

## Verdict

**Approve.** The original one-liner was correct for what it caught, and closing the lightning double-tap is worth doing. It just didn't reach the on-chain case it claimed, and turning the spinner into a latch made an existing gap in `.paymentFailed` user-visible in a bad way. Both are addressed on the branch now; what's left is the manual pass above, which matters more than usual here given one of these paths moves real money.

PR #49, opened 2026-07-21
Replace stringly-typed cache dispatch with a Codable model

tommelters:
The home-screen display cache (balance, fiat values, block height, and a transactions snapshot) was read/written through
updateCachedData(data:Any,key:String) / getCachedData(key:)->Any? with if-key ladders duplicated across branches, forcing an as? cast at every call site.

Replace it with a Codable WalletCache + CachedTransaction model and typed accessors (cachedSatsBalance, cachedEurValue, cachedHeight, cachedHomeTransactions, ...). The cache is one JSON value mutated through a single serial queue, so concurrent writers can no longer interleave a read-modify-write and drop a field. Legacy NSDictionary caches are migrated once on first access, so upgrading users keep their instant/offline home screen. All 16 call sites updated.

Move CacheManager into a new Cache/ folder alongside WalletCache.swift (which holds the model plus the CacheManager cache extension).


RubenWaterman (commented):
Reviewed this and rebased it onto current `android` to clear the conflicts. No code changes beyond the rebase — everything below is low-severity and recorded for context rather than as a blocker.

Same stacking as #45, #47 and #48: forked at `f2e17d3`, carrying stale copies of #42, #45 and #47. The real work is one commit, `b5335db`.

`git rebase --onto origin/android e13bc59` replayed it with **zero conflicts**, including the `CacheManager.swift` rename into `Cache/`.

```
before:  22 commits, +1077/-500 across 27 files, CONFLICTING/DIRTY
after:    1 commit,   +219/-182 across 10 files, MERGEABLE/CLEAN
```

Because this commit rewrites *and moves* `CacheManager.swift` — a file that gained code in #46 — I checked explicitly that nothing merged since the fork was swallowed:

- `deleteTransactionNote` (#46)
- `runClaimedNodeStart`, `isContinuingStartWallet`, `FeeEstimates`, `wasRemovingWalletForIncorrectPin` (#45)
- `bdkFullScanTimedOut`, the light-sync guard, `onchainsynctimedout` (#47)
- both `confirmSpinner.isAnimating` guards and the `.paymentFailed` fix (#48)

All present. Build succeeds. New SHA `9a2b0e3`.

## Overview

Replaces `updateCachedData(data:Any,key:String)` / `getCachedData(key:)->Any?` — two duplicated if-key ladders, ~166 lines, an `as?` cast at every call site — with a `Codable WalletCache` + `CachedTransaction` and typed accessors, all mutations funnelled through one serial queue. Legacy `NSDictionary` caches migrate once on first access.

## What's right

- **The core claim holds.** Every read-modify-write now goes through `cacheQueue.sync`, so two writers can't interleave and drop a field — which the old code genuinely could, since each `updateCachedData` did its own `defaults.value` → `mutableCopy` → `set`.
- **Call sites are faithful 1:1 translations.** Diffed all of them; no logic changed, only the accessor.
- **No round-trip data loss.** Compared the old `parseTransactions` key set against `CachedTransaction`'s fields — identical, modulo casing (`isswap`→`isSwap`, `onchainid`→`onchainID`, …). Nothing the old cache persisted is dropped.
- **No re-entrancy deadlock.** `loadCacheLocked` calls `getTransactions` while holding `cacheQueue`; `getTransactions` touches no cache accessor, so it's safe today. Worth a comment there, since a future edit that reads a `cached*` property would deadlock the app silently.
- **Migration deletes the legacy key** after a successful write, so it runs exactly once.

## Findings

### Medium — every read and write decodes/encodes the entire cache

`readCache()` fully JSON-decodes `WalletCache` — including the whole transactions array — on *each* property get. `mutateCache` decodes, mutates, re-encodes and writes on each set.

That lands on the hot paths:

- `ShowCachedData.showCachedData()` — 4 reads = 4 full decodes, on the cold-start path this cache exists to make fast.
- `LoadWalletData` — writes at `:123`, `:297`, `:306`, `:356`, `:357`; each re-encodes every cached transaction.

The old code was arguably cheaper: `UserDefaults` keeps the plist-decoded `NSDictionary` in memory, so reads were dictionary lookups rather than a JSON codec run. The refactor is a clear correctness and readability win, but it may be a small performance regression on the exact screen it serves.

A memoized `static var` behind the same queue, invalidated on write, would fix it without changing the API.

### Low — a decode failure silently discards the cache, and the schema is brittle

If `JSONDecoder` fails and no legacy dictionary is present, `loadCacheLocked` returns an empty `WalletCache()`, and the next `mutateCache` writes that empty value over the undecodable data.

Every field in `CachedTransaction` is non-optional except `height`, so **adding one non-optional field later breaks decoding for every existing user** — cache wiped, blank home screen until the first sync completes. Blast radius is cosmetic since this is a display cache, but it's worth a comment on the struct, a `version` field, or a rule that new fields are always optional.

### Low — undocumented behaviour change: `.filter { $0.timestamp != 0 }`

`cachedHomeTransactions`' getter filters out zero-timestamp transactions. The old `getCachedData(key: "transactions")` did not, and the description doesn't mention it.

In practice it probably drops nothing real: confirmed on-chain uses the block timestamp, unconfirmed uses `latestUpdateTimestamp`, and lightning goes through `getInvoiceTimestamp`, which never returns 0. So it most likely just cleans malformed legacy entries — plausibly the intent.

But it's read-side only, so a zero-timestamp entry is still *written* and simply never comes back — invisible either way. Filtering on write, or dropping the filter, would be more honest than a silent read-side discard.

### Low — `WalletCache.balance` is dead

Declared in the struct and populated by the migration, but there's no `cachedBalance` accessor and no call site reads it. The old dispatch had the same unused `"balance"` key (call sites only ever used `"satsbalance"`), so this isn't a regression — the refactor carries the dead field forward. Easy to drop.

### Low — unknown `swapStatus` decodes to `.succeeded`

```swift
case "pending": .pending
case "failed":  .failed
default:        .succeeded
```

The encoder only ever writes those three strings, so `default` is reachable only via corruption or a future rename. Still, `.succeeded` is the least safe fallback for a swap — it tells the user funds arrived. `.pending` or `.failed` fails safe.

### Nits

- **Inconsistent enum encoding** — `swapStatus` persists as a readable string, `swapDirection` as a magic `0`/`1`. Making both `Codable`, or both strings, would be more consistent and leave the stored JSON self-describing.
- **`readCache()` mutates** on first access (migrates, writes, deletes the legacy key). Intended, but the name hides it.
- **`import UIKit`** in `WalletCache.swift` is only needed for `CGFloat`; `Foundation` + `CoreGraphics` would do.

## Testing

No tests — and this is the first PR in the series where that reads as a genuine miss rather than a scope question. `CachedTransaction` and the migration are **pure value transforms with no LDK, BDK, or network dependency**:

- `CachedTransaction(t).toTransaction()` round-trips every field — the exact property verified by hand above, which a test would pin permanently.
- `loadCacheLocked`'s legacy migration is a pure function of an `NSDictionary`.
- The `swapStatus` / `swapDirection` mappings are three and two cases.

Roughly 40 lines of test for the highest-value guarantee in the PR. It would still mean establishing the first real test in `bittrTests`, so it stays a judgement call — but the argument is much stronger here than on #45–#48.

Meanwhile the branch is **compile-verified only**; none of this has been exercised at runtime. Worth a manual pass:

- **upgrade path** — install the old build, use it so a legacy cache exists, then upgrade: the home screen should still show cached balance and transactions instantly, and the old `cache` key should be gone
- **fresh install** — no legacy cache, no crash, cache populates
- cold start with cached data in airplane mode — balance, conversion and history all render

## Verdict

**Approve.** A real improvement: it deletes a genuinely error-prone construct, the concurrency claim is sound, and the migration means upgrading users keep their instant home screen. Nothing here is blocking. The one finding worth a second look eventually is the per-access codec cost, precisely because this cache exists for startup speed — but it's a small regression on a path that was never fast to begin with, not a reason to hold the PR.

PR #50, opened 2026-07-21
Refactor performSwapMatching into a dedicated, type-safe file

tommelters:
Move performSwapMatching out of LoadWalletData into SwapMatching.swift and rewrite it off a typed [String: [Transaction]] instead of an NSMutableDictionary with ~10 force-casts. 

The 172-line method becomes a small orchestrator plus handleCompletedSwap / makeCompletedSwapTransaction / handlePendingSwap helpers; grouping order, reference-mutation of suggested swaps, and the combined-swap accounting are all preserved.

Also drop the redundant argument label on storeLightningTransaction (storeLightningTransaction(_:)) and update its call sites.

### DIFF
--- a/ios/bittr.xcodeproj/project.pbxproj
+++ b/ios/bittr.xcodeproj/project.pbxproj
@@ -67,6 +67,7 @@
 		053C37542FD28E8E00A510DF /* BittrArticles.swift in Sources */ = {isa = PBXBuildFile; fileRef = 053C37532FD28E8800A510DF /* BittrArticles.swift */; };
 		054D42DF2F7B739E00599720 /* AnimationViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 054D42DE2F7B739800599720 /* AnimationViewController.swift */; };
 		05512F26300F80AF00E75272 /* WalletCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05512F25300F80AC00E75272 /* WalletCache.swift */; };
+		05512F28300F957C00E75272 /* SwapMatching.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05512F27300F957900E75272 /* SwapMatching.swift */; };
 		0554035D2CB7EA04006F27CA /* LNURLDecoder in Frameworks */ = {isa = PBXBuildFile; productRef = 0554035C2CB7EA04006F27CA /* LNURLDecoder */; };
 		0554035F2CB82B2E006F27CA /* ReceiveVCLanguage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0554035E2CB82B28006F27CA /* ReceiveVCLanguage.swift */; };
 		055403692CC14FC5006F27CA /* SendLNURL.swift in Sources */ = {isa = PBXBuildFile; fileRef = 055403682CC14FC0006F27CA /* SendLNURL.swift */; };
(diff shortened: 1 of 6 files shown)


RubenWaterman (commented):
Reviewed this, rebased it onto current `android` to clear the conflicts, and pushed a second commit (`409c6ab`) with the cleanups. Findings are recorded as they stood against `5a057f7`, with what changed noted under each.

Sixth PR in a row forked from `f2e17d3`, this one stacked on the pre-rebase #49. The real work is one commit, `5a057f7`.

`git rebase --onto origin/android b5335db` replayed it with **zero conflicts**. Verified nothing merged since the fork was lost: `runClaimedNodeStart` (#45), `deleteTransactionNote` (#46), `bdkFullScanTimedOut` (#47), `confirmSpinner.isAnimating` (#48), `WalletCache` / `cachedHomeTransactions` (#49) all present.

```
before:  23 commits, +1246/-682 across 30 files, CONFLICTING/DIRTY
after:    2 commits,  +173/-182 across  6 files, MERGEABLE/CLEAN
```

## Overview

Moves the 172-line `performSwapMatching` out of `LoadWalletData.swift` into `SwapMatching.swift`, replacing an `NSMutableDictionary` and ~10 force-casts with a typed `[String: [Transaction]]`, split into an orchestrator plus three helpers. Also drops the argument label on `storeLightningTransaction(_:)`.

## Correctness — it is faithful

Compared line by line against the original. The accounting is the part that matters, and it's preserved exactly:

- **Grouping** — same key (`lnDescription`), same append order within a group.
- **Completed-swap early exits** — the old `if / else-if / else` maps cleanly onto the two guarded `return`s. `removeAll { $0.id == X }` matches the old `enumerated().reversed()` removal, including removing *all* matches rather than the first.
- **Combined-transaction accounting** — `sent`, `fee`, direction, `isLightning`, the per-leg loop (`isSwap ? leg.lightningID : leg.id`, height, timestamp, the `received - sent` / `sent - received` asymmetry) and the both-onchain failed-swap refund block are all equivalent.
- **Pending swaps** — the suggested-swap branch still mutates the `Transaction` **by reference** so the change lands in the array; `return` from the helper correctly replaces the old `continue`.
- **Group iteration order** — both `NSMutableDictionary` and Swift `Dictionary` are unordered, so this isn't a change, and it doesn't matter: both call sites (`LoadWalletData:115`, `HomeViewController:319`) sort by timestamp immediately after.
- **Label drop** — all 8 `storeLightningTransaction` call sites updated; none missed.

A genuine readability win with no behavioural drift.

## Findings

### Low — the fee-summing rationale comment was deleted

The original carried six lines explaining *why* both legs' fees are summed: the home row renders cost as `(received - sent - fee)`, so dropping either side's fee understates the displayed swap cost, and inbound legs have `fee = 0` so summing both is safe. The refactor kept `swapTransaction.fee = first.fee + second.fee` under a bare `// Amount and fees.`

That comment documents an accounting bug that was fixed once already, and the `fee = 0` invariant is exactly what stops a future reader seeing the sum as a double-count and "correcting" it.

**Restored** in `409c6ab`.

Worth noting this is the third PR in the series where a substantive rationale comment was dropped during a refactor (#45's force-close notes, #47's watchdog explanation). A mechanical refactor should carry comments across untouched — they're the part that can't be re-derived from the code.

### Low — the helpers were instance methods that never used `self`

`handleCompletedSwap`, `makeCompletedSwapTransaction` and `handlePendingSwap` were declared on `extension [Transaction]` but none read the receiver — they operate purely on their parameters plus the `inout` array. As written, every `[Transaction]` in the app gained three methods with nothing to do with its contents.

**Fixed** in `409c6ab` — now `private static func`, called via `Self.`. Purely structural; no behaviour change.

### Nits — all addressed in `409c6ab`

- **`import UIKit`** → `Foundation`; nothing here needs UIKit.
- **`print("Swap ID: …")`** folded into the adjacent `Log.info` calls, so the swap ID still lands in the logs without a stray `print` in production.

### Nit — argument-label drop, left as-is

`storeLightningTransaction(_:)` reads fine, but it now differs from its neighbours (`storeTransactionNote(txid:note:)`, `getSwapID(dateID:)`, `getSuggestedSwapStatus(dateID:)`), which all label their parameters. Deliberately **not changed** — dropping the label is a stated goal of this PR, and this is a balanced trade rather than a defect. Flagging only so the inconsistency is a known choice.

### Observation — groups of 3+ are still silently ignored

`count == 2` and `count == 1 && !isSwap` are handled; anything else falls through and both legs stay as separate rows. Unchanged from the original, so not a regression — but the typed rewrite makes it visible in a way the old force-cast ladder didn't, and a retried or duplicated swap sharing an `lnDescription` would land there. Worth a comment saying it's deliberate, or a `Log.info` so it's diagnosable if it ever happens. Left alone as out of scope.

## Testing

No tests, and this is the strongest candidate in the series so far: `performSwapMatching` is a **pure function on `[Transaction]`** — no LDK, no BDK, no network, with `CacheManager.getSwapID` / `getSuggestedSwapStatus` the only external reads.

Given it computes the swap amounts and fees shown to users, a handful of table-driven cases (completed L→O, completed O→L, both-onchain failed refund, pending, suggested) would lock in exactly the accounting this PR asserts it preserved. That equivalence was verified here by reading; a test would keep it true.

The branch is **compile-verified only — nothing has been exercised at runtime.** Worth a manual pass:

- a completed lightning→onchain swap: amount and fee match pre-refactor
- a completed onchain→lightning swap
- a failed swap (both legs on-chain) — refund direction picks the right `onchainID` / `lightningID`
- a pending swap, and a suggested swap (confirm it stays flagged in the list)

## Verdict

**Approve.** The best-scoped PR in this series: a 172-line method with ~10 force-casts becomes four readable functions, and the behaviour genuinely is preserved — including the parts that move money on screen. Everything found was cosmetic apart from the deleted fee comment, and that's back on the branch. What remains is the manual pass above.

PR #51, opened 2026-07-22
Refactor maximum sendable onchain satoshis in SendVC and SwapVC

tommelters:
- In the **SendVC** and **SwapVC**, the estimation of **maximum sendable onchain funds** was clumsy and error-prone. This has been refactored.
- Maximum possible drain is now calculated using **BDK's built-in drainWallet() function**.
- For estimates where no address is yet known, estimations are done based on heaviest possible address.
- All calculations now guard for a **minimum amount of 0 satoshis**, and a **minimum fee of 1 sat/Vbyte**.
- Outcomes guard against LDKNode's own **listBalances().spendableOnchainBalanceSats** number.
- **Up-to-date fee rates** now get **downloaded immediately** upon opening SendVC or SwapVC (instead of after clicking Next).
- When the user intends to drain their onchain funds, we now broadcast the transaction using **LDKNode's built-in sendAllToAddress()** function. So the transaction won't fail, as the sendable amount is calculated by LDKNode automatically in combination with the fee rate that the user has selected.

### DIFF
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -9,6 +9,12 @@ import Foundation
 import BitcoinDevKit
 import Sentry
 
+struct OnchainDrainPreview {
+    let sendableSats:UInt64
+    let feeSats:UInt64
+    let vsize:UInt64
+}
+
 extension BitcoinManager {
     
     func didStartBDK() -> Bool {
@@ -271,12 +277,74 @@ extension BitcoinManager {
         return true
     }
     
-    func getSize(address:String, amountSats:Int) throws -> UInt64 {
+    func getSize(address:String, amountSats:Int, selectedVbyte:Float?) throws -> UInt64 {
         
-        let tx = try self.getTx(address: address, amountSats: amountSats, selectedVbyte: nil)
+        let tx = try self.getTx(address: address, amountSats: amountSats, selectedVbyte: selectedVbyte)
         let size = tx.vsize()
         
         return size
+    }
+    
+    // MARK: - Onchain maximum
+    
+    // Get a placeholder output script to calculate fees before knowing the recipient.
+    static var largestCommonOutputScript: Script {
+        Script(rawOutputScript: [0x51, 0x20] + [UInt8](repeating: 0, count: 32))
+    }
+    
+    // Get maximum sendable onchain transaction.
+    func previewOnchainDrain(toScript script:Script, satPerVb:UInt64) throws -> OnchainDrainPreview {
+        guard let wallet = self.bdkWallet else {
+            throw WalletError.walletNotInitiated
+        }
+        
+        // Minimum 1 sat/vB.
+        let feeRate = try FeeRate.fromSatPerVb(satVb: max(satPerVb, 1))
+        
+        let psbt = try TxBuilder()
+            .drainWallet()
+            .drainTo(script: script)
+            .feeRate(feeRate: feeRate)
+            .finish(wallet: wallet)
+        let _ = try wallet.sign(psbt: psbt, signOptions: nil)
+        let tx = try psbt.extractTx()
+        
+        // drainTo produces exactly one output, so its value is the maximum.
+        guard let drainOutput = tx.output().first else {
+            throw WalletError.walletNotInitiated
+        }
+        
+        return OnchainDrainPreview(sendableSats: drainOutput.value, feeSats: try psbt.fee(), vsize: tx.vsize())
+    }
+    
+    // Calculate the maximum sendable onchain amount, with or without known recipient.
+    func maximumSendableOnchainDrain(toAddress address:String?, satPerVb:UInt64) throws -> OnchainDrainPreview {
+
+        let script:Script
+        if let address = address,
+           let parsed = try? Address(address: address, network: EnvironmentConfig.bitcoinDevKitNetwork) {
+            // Recipient is known.
+            script = parsed.scriptPubkey()
+        } else {
+            // Recipient is unknown.
+            script = BitcoinManager.largestCommonOutputScript
+        }
+        
+        let preview = try self.previewOnchainDrain(toScript: script, satPerVb: satPerVb)
+        
+        // Verify that LDKNode agrees with BDK's maximum sendable onchain amount.
+        let spendable = UInt64(max(BitcoinManager.shared.bittrWallet.satoshisOnchainSpendable, 0))
+        guard spendable > 0, preview.sendableSats + preview.feeSats > spendable else {
+            // LDKNode's maximum sendable onchain amount is lower than BDK's.
+            // Therefore, stick with LDKNode as the source of truth.
+            return preview
+        }
+        let clamped = spendable > preview.feeSats ? spendable - preview.feeSats : 0
+        return OnchainDrainPreview(sendableSats: clamped, feeSats: preview.feeSats, vsize: preview.vsize)
+    }
+    
+    func maximumSendableOnchainSats(toAddress address:String?, satPerVb:UInt64) throws -> UInt64 {
+        return try self.maximumSendableOnchainDrain(toAddress: address, satPerVb: satPerVb).sendableSats
     }
     
     func getTx(address:String, amountSats:Int, selectedVbyte:Float?) throws -> BitcoinDevKit.Transaction {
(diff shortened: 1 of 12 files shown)


RubenWaterman (commented):
Reviewed this, rebased it onto current `android`, and pushed two commits addressing what came out of it. Findings are recorded as they stood against `fe77429`, with what changed noted under each.

## Conflicts — resolved

Seventh PR forked from `f2e17d3`. The real work is one commit, `fe77429`.

**One genuine textual conflict**, in `LoadWalletData.swift`: this PR adds a `balances` local and `spendableOnchainBalanceSats` at the same spot where #45 hoisted `guard let node` to the top of the function. Resolved by keeping #45's hoist and taking the new lines; the PR's re-declared inline guard was dropped as redundant.

**And a semantic clash a textual merge cannot see.** `fetchFeeEstimatesThenSetSendAllLabel` is new code written against `getFeeEstimates()`'s old `NSDictionary` return, subscripted by key. #45 changed that API to return a `FeeEstimates` struct, so once rebased the branch didn't compile:

```
error: value of type 'FeeEstimates' has no subscripts
```

Fixed in `615a022`, matching how #45 rewrote the other four call sites. Worth dwelling on: the `CONFLICTING` flag would never have caught this, and neither would a clean-looking rebase — it only surfaced because the branch was built.

```
before:  24 commits, +1456/-755 across 36 files, CONFLICTING/DIRTY
after:    3 commits,  +266/-77  across 12 files, MERGEABLE/CLEAN
```

Verified all merged work survived: #45's hoisted guard, #46, #47, #48, #49's `WalletCache`, #50's static helpers and fee comment.

## Overview

Replaces the old "build a transaction you expect to fail, then parse the error" approach to computing the maximum sendable on-chain amount with a real BDK `drainWallet()` PSBT, and switches the actual drain broadcast to LDK's `sendAllToAddress(retainReserve: true)`.

## What's genuinely good

- **Deleting the error-string parsing is the highlight.** The old code did `String(error.localizedDescription.split(separator: " ")[7]).toNumber()` — an index-out-of-range crash waiting on any wording change, and locale-dependent besides. Building a real drain PSBT and reading `output().first.value` is the correct approach.
- **`sendAllToAddress(retainReserve: true)`** is the right API for a drain. LDK computes the amount against the actual fee rate and keeps the anchor-channel reserve, so the transaction can't fail for being a few sats over.
- **`largestCommonOutputScript`** — `[0x51, 0x20] + 32 zero bytes` is a 34-byte P2TR script, tying P2WSH as the heaviest standard output. Estimating against it overestimates the fee and so *underestimates* the sendable amount: conservative in the safe direction.
- **`getSize` now takes the selected fee rate** instead of always passing `nil`. Since fee rate affects coin selection and therefore input count, the old size estimate could be wrong for the rate actually used.
- Minimum 1 sat/vB and floor-at-zero guards are applied consistently across the new paths.

## Findings

### High — the confirmation screen could show one amount and send another

```swift
} else {
    self.isSendingMaximum = drain != nil && self.onchainAmountInSatoshis >= Int(drain!.sendableSats)
}
```

The `didTapAvailable` branch above it corrected `onchainAmountInSatoshis` **and** `confirmSatoshis` to the drain amount. This branch set the flag and corrected neither.

`checkSendOnchain` only rejects amounts above `satoshisOnchainSpendable`, while the drain maximum sits roughly a fee below that. So any amount in the gap passed the guard, tripped `>=`, and became a full drain:

- spendable = 100,000, fee ≈ 500 → drain max = 99,500
- user types **99,800** → guard passes, `99,800 >= 99,500` → `isSendingMaximum = true`
- confirmation alert reads **"send 99,800 satoshis"** (it interpolates `confirmSatoshis`)
- `performOnchainTransaction` sees the flag, calls `sendAllOnchainPayment`, and LDK broadcasts **≈99,500**

The window is about one fee's worth, so no funds are at risk — it's the user's own drain. But a Bitcoin send confirmation that names an amount the wallet won't broadcast is a trust problem, not a rounding detail.

**Fixed** in `c668bc2` — both drain paths now restate the amount as the drain figure, so the screen always states what will be sent.

### Medium — the swap screen skipped the LDK clamp the description promises

The description says outcomes "guard against LDKNode's own `listBalances().spendableOnchainBalanceSats`". That guard lives in `maximumSendableOnchainDrain`, but `SendableAmount.swift` called **`previewOnchainDrain` directly**.

BDK knows nothing about the reserve LDK Node holds back for anchor channels, so the swap screen could quote a maximum that includes reserved funds while the send screen clamped correctly. Two screens, two answers, same wallet.

**Fixed** in `c668bc2` — SwapVC now goes through `maximumSendableOnchainDrain` with `toAddress: nil`, which still quotes against the heaviest common output script since the Boltz address isn't known at quote time.

### Medium — the clamp failed open when spendable was zero

```swift
guard spendable > 0, preview.sendableSats + preview.feeSats > spendable else {
    return preview
}
```

With `spendable == 0` the guard exited early and returned BDK's **unclamped** drain. The intent was presumably "balance not loaded yet, don't quote zero", but it conflated *not yet loaded* with *genuinely nothing spendable* — and the latter is exactly the case where clamping matters most, everything reserved for anchor channels. `satoshisOnchainSpendable` defaulted to `0` and is only populated by `loadWalletData`, so the not-loaded state was real and reachable.

**Fixed** in `c668bc2` by making `satoshisOnchainSpendable` an `Int?`: nil means not loaded and skips the clamp, zero clamps to zero. That touched 8 call sites — wider than ideal, but a sentinel value or a parallel "loaded" flag would hide exactly the distinction the bug was about.

### Low — `UInt64(selectedVbyte)` converted before it clamped

```swift
let feeRateSatVb = max(UInt64(selectedVbyte), 1)
```

`selectedVbyte` is a `Float`, and `UInt64(_:)` **traps** on a negative or NaN value — the `max(…, 1)` ran afterwards and couldn't guard anything. Not just a missing clamp: a crash on a line that reads as though it's already protected. Reachable only if a fee rate goes negative or NaN, which current sources shouldn't produce.

**Fixed** in `c668bc2` — `UInt64(max(selectedVbyte, 1))`, matching what the new code already does correctly elsewhere.

### Low — the quote used the medium rate while the send used the selected one

`maximumSendableOnchainDrain` is called with `feePerVbMedium`, but the user then picks low/medium/high on the confirm screen and `sendAllToAddress` recomputes at that rate. Choosing "high" meant less was sent than was quoted.

**Fixed** in `c668bc2` — the drain total is carried on `SendViewController` and `switchToFee` restates the amount whenever the rate changes.

One caveat worth knowing: that restatement is arithmetic (`drainTotal − selectedFeeInSats`), not a re-quote. Rebuilding and re-signing the PSBT on every fee tap would put the work straight back on the main thread it was just moved off. It tracks LDK closely because it reuses the same vsize, but `sendAllToAddress` does its own coin selection and reserve retention, so the displayed figure can still differ slightly from what lands on chain. Exact agreement isn't reachable without a quote API LDK doesn't expose.

### Low — every estimate built and signed a PSBT on the main thread

`previewOnchainDrain` runs `TxBuilder().drainWallet()…finish()`, then `wallet.sign()`, then `extractTx()`. Signing is needed for an accurate vsize, so the work is justified — but `setSendAllLabel` called it on the main thread, and this PR moves fee fetching to screen-open so it now runs more often. On a wallet with many UTXOs that's synchronous crypto on the UI thread.

**Fixed** in `c668bc2` — moved to a background queue with the spinner held until a number is ready. `SendableAmount` already did its equivalent inside a `Task`.

### Nit

`previewOnchainDrain` threw `WalletError.walletNotInitiated` when `tx.output().first` was nil. The wallet *is* initiated at that point — the drain just produced no output. **Fixed** in `c668bc2` with a distinct `drainProducedNoOutput` case, so the Sentry signal means something.

## Testing

No tests. `previewOnchainDrain` needs a funded BDK wallet so it isn't unit-testable as written — but `maximumSendableOnchainDrain`'s clamp arithmetic is pure given a preview and a spendable figure, and it's where two of the findings above lived. Extracting that comparison would make the zero-spendable and over-max cases directly testable.

The branch is **compile-verified only — none of this has been exercised at runtime.** Since this path drains real wallets, it wants a manual matrix before merge:

- tap "send all" and confirm the quoted number matches what is actually broadcast
- type an amount a few hundred sats **above** the quoted max — the High finding; check the confirmation against what lands on chain
- with an open anchor channel, confirm the quote excludes the reserve on **both** the send and swap screens
- select low / medium / high after tapping "send all" and check the restated amount each time
- a wallet with zero spendable but non-zero total
- a fee-estimate fetch failure → the fallback quotes the spendable balance

## Verdict

**Approve.** The direction is right and the old error-parsing code deserved to go. The one thing that would have blocked it — a confirmation screen that could name an amount the wallet wouldn't send — is fixed on the branch, along with the two clamp gaps and the trapping conversion. What remains is the manual pass above, which matters more here than on any PR in this series so far.

PR #52, opened 2026-07-22
Quote lightning-to-onchain swap fees as a range

tommelters:
**Problem**
When performing a lightning-to-onchain swap, the quoted **expected fee** was usually much higher than the eventual **actual fee**. 
This is due to lightning routing fees turning out lower than the potential maximum.

**Solution**
Same as in ConfirmSendVC, **show expected fees as a range**.
For lightning-to-onchain swaps:
- the **alert** now says "The expected fee to move X satoshis (€ X) is between X and X satoshis (EUR X - X). Do you wish to proceed?" 
- the **SwapStatusVC** says "X - X satoshis".

### DIFF
--- a/ios/bittr/Extensions/String.swift
+++ b/ios/bittr/Extensions/String.swift
@@ -19,25 +19,31 @@ extension String {
     }
     
     func addSpaces() -> String {
-        
+
         var balanceValue = self
         if balanceValue.fixDecimals().contains(Locale.current.decimalSeparator!) {
             balanceValue = String(self.fixDecimals().split(separator: Locale.current.decimalSeparator!)[0])
         }
-        
+
+        // Group with a non-breaking space. A plain space is a legal line-break
+        // opportunity, so a wrapping label could split "1 213" into "1" and
+        // "213" on separate lines and read as two numbers. Output is display
+        // only — no call site parses it back — so the narrower space is safe.
+        let separator = "\u{00A0}"
+
         switch balanceValue.count {
         case 4:
-            balanceValue = balanceValue[0] + " " + balanceValue[1..<4]
+            balanceValue = balanceValue[0] + separator + balanceValue[1..<4]
         case 5:
-            balanceValue = balanceValue[0..<2] + " " + balanceValue[2..<5]
+            balanceValue = balanceValue[0..<2] + separator + balanceValue[2..<5]
         case 6:
-            balanceValue = balanceValue[0..<3] + " " + balanceValue[3..<6]
+            balanceValue = balanceValue[0..<3] + separator + balanceValue[3..<6]
         case 7:
-            balanceValue = balanceValue[0] + " " + balanceValue[1..<4] + " " + balanceValue[4..<7]
+            balanceValue = balanceValue[0] + separator + balanceValue[1..<4] + separator + balanceValue[4..<7]
         case 8:
-            balanceValue = balanceValue[0..<2] + " " + balanceValue[2..<5] + " " + balanceValue[5..<8]
+            balanceValue = balanceValue[0..<2] + separator + balanceValue[2..<5] + separator + balanceValue[5..<8]
         case 9:
-            balanceValue = balanceValue[0..<3] + " " + balanceValue[3..<6] + " " + balanceValue[6..<9]
+            balanceValue = balanceValue[0..<3] + separator + balanceValue[3..<6] + separator + balanceValue[6..<9]
         default:
             balanceValue = balanceValue[0..<balanceValue.count]
         }
--- a/ios/bittr/Language.swift
+++ b/ios/bittr/Language.swift
@@ -330,6 +330,7 @@ class Language: NSObject {
             "nodatareceived": "No data was received",
             "swapamountexceeded": "You can only move up to <amount> satoshis at a time. Please enter an amount within this limit.",
             "swapfunds3": "The expected fee to move <b><amount> satoshis</b> (<convertedamount>) is <b><feesamount> satoshis</b> (<convertedfees>).",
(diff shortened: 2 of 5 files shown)


RubenWaterman (commented):
Reviewed this and rebased it onto current `android` to clear the conflicts. No code changes beyond the rebase.

Eighth PR forked from `f2e17d3`, stacked on the pre-rebase #50. The real work is one commit, `68287db`.

`git rebase --onto origin/android 5a057f7` replayed it with **zero conflicts**, and it **builds clean**. I checked the build specifically: #51 rebased just as cleanly and then failed to compile, because it used an API #45 had changed — a semantic clash no textual merge can see. This PR touches swap fee display, which #51 also touched, so it was the obvious candidate for a repeat. It isn't.

Verified all merged work survived: #51's `satoshisOnchainSpendable`, `drainTotalSats`, `maximumSendableOnchainDrain` and `drainProducedNoOutput`, plus #46–#50.

```
before:  24 commits, +1298/-692 across 32 files, CONFLICTING/DIRTY
after:    1 commit,    +52/-10  across  4 files, MERGEABLE/CLEAN
```

## Overview

Two changes, only one of which the description mentions:

1. **Fee range** (described) — L→O swaps quote "between X and Y satoshis", since the routing fee isn't known until the payment finds a route. Five computed properties on `Swap` plus a new `swapfunds3range` string.
2. **Claim-fee double-count fix** (in the commit subject, *not* in the PR body) — the claim transaction fee is no longer added to the L→O total.

## The undisclosed fix is correct — verified

This is the consequential half, so I traced it rather than trusting the subject line.

For a lightning→onchain swap:

- `SwapManager.swift:396` — `onchainAmountWithFee = amountSat + (claimTransactionFee ?? 0)`. The claim fee is added to the amount **requested from Boltz**.
- `SwapManager.swift:697` — `onchainFees = invoiceAmount - finalOnchainAmount`. Since the invoice covers the requested amount, that spread **already contains the claim fee**.
- The old code then added `+ (claimTransactionFee ?? 0)` on top of it.

The old implementation carried the contradiction in its own comment — *"the claim transaction fee is included in the on-chain amount so the user receives exactly what they input"* — directly above the line adding it a second time. `SwapManager.swift:699` says the same. The new `definiteFees`, returning `onchainFees` alone for L→O, matches what the code actually does.

Checked the other direction too: `SwapManager.swift:294-301` never sets `claimTransactionFee` for O→L, so that branch's total is unchanged.

**A bonus the PR doesn't claim:** the `catch` fallback at `SwapManager:718` doesn't set `claimTransactionFee`, so previously the same swap quoted a *different* fee depending on whether the claim-fee call succeeded. Both paths now produce the same number.

## Findings

### Medium — the description omits the fee-amount change

The body describes a presentation change: show fees as a range. The commit also **lowers the quoted fee for every L→O swap** by the claim transaction fee (99 vbytes × the fastest rate — tens to hundreds of sats depending on the mempool).

That's a correctness fix, and a good one. But a reviewer reading only the description would approve a formatting change and merge a change to the numbers users see before committing funds. Worth adding to the body so the next reader knows what landed, and so it's findable if the fee display is ever questioned.

### Low — "between X and X" is reachable

```swift
var hasVariableFee:Bool { maximumRoutingFee > 0 }
var minimumTotalFees:Int { definiteFees + (hasVariableFee ? 1 : 0) }
```

If `lightningFees == 1`, `hasVariableFee` is true but `minimumTotalFees == maximumTotalFees`, so the alert renders *"between 500 and 500 satoshis"* and `SwapStatusVC` shows *"500 - 500 sats"*.

The fiat side already guards this (`convertedMinimum == convertedMaximum ? convertedMaximum : …`); the sats side doesn't. Either apply the same collapse in `formattedTotalFees()` and the message selection, or make `hasVariableFee` require `> 1`.

### Low — the 1-sat floor is an unstated assumption

`minimumTotalFees` adds exactly 1 sat in the variable case, i.e. assumes routing can't be free. Over a direct zero-fee hop the true minimum is `definiteFees`. Harmless for a quote, and it does avoid a degenerate range — but it's a guess baked into a displayed number with nothing saying so.

### Nits

- **`claimTransactionFee` no longer feeds any displayed total.** L→O excludes it (correctly); O→L never sets it. So `Swap.swift:51`'s `+ (self.claimTransactionFee ?? 0)` is permanently `+ 0` — dead arithmetic implying a coupling that doesn't exist. Drop it, or note why it's kept.
- **`totalFees` in the swap file is write-only.** `updateSwapFileWithFees` sets it and nothing reads it, so switching it to `maximumTotalFees` has no user-visible effect. Third dead field in this series after #46's `Transaction.note` and #49's `WalletCache.balance` — might be worth a sweep.
- **`formattedTotalFees()` is a method** while its four siblings are computed properties. Trivial consistency thing.

## Testing

No tests — and this is the easiest case in the series so far. The five new members are **pure `Int` arithmetic on a value type**: no LDK, no BDK, no network, no `CacheManager`. Five properties, five assertions, and the "between X and X" case above would have been caught by writing them.

The branch is **compile-verified only**; none of this has been exercised at runtime. Manual checks:

- L→O quote shows a range, and `SwapStatusVC` shows the same figures
- O→L still shows a single value
- force the claim-fee call to fail → the quote should now match the success path
- the actual fee on completion falls inside the quoted range
- a swap whose maximum routing fee is 1 sat (the degenerate range)

## Verdict

**Approve.** Small, correct, and the undisclosed half is the better half — it fixes a real overstatement in what users are told a swap costs, and makes two previously divergent code paths agree. I'd ask for the PR body to mention the claim-fee fix before merge; the "between X and X" collapse is a two-line follow-up.

PR #53, opened 2026-07-22
Minor refactor to Swap file

tommelters:
toDictionary and toSwap were 142 lines of near-identical blocks, one pair per field. NSMutableDictionary.setValue already removes the key when handed a nil value, so the `if x != nil { setValue(x!) }` guards were redundant; and every optional property starts out nil on a fresh Swap, so `if let x = d[k] as? T { swap.x = x }` is just `swap.x = d[k] as? T`. Both collapse to one line per field, grouped to match the property declarations.

### DIFF
--- a/ios/bittr/Cache/WalletCache.swift
+++ b/ios/bittr/Cache/WalletCache.swift
@@ -66,7 +66,6 @@ extension CacheManager {
             return WalletCache()
         }
         var cache = WalletCache()
-        cache.balance = legacy["balance"] as? String
         cache.satsBalance = legacy["satsbalance"] as? String
         cache.conversion = legacy["conversion"] as? String
         cache.eurValue = legacy["eurvalue"] as? CGFloat
@@ -88,7 +87,6 @@ extension CacheManager {
 }
 
 struct WalletCache: Codable {
-    var balance: String?
     var satsBalance: String?
     var conversion: String?
     var eurValue: CGFloat?
--- a/ios/bittr/Swaps/Swap Manager/SwapManager.swift
+++ b/ios/bittr/Swaps/Swap Manager/SwapManager.swift
@@ -510,7 +510,6 @@ class SwapManager: NSObject {
                             swapVC.thisSwap!.claimLeafOutput = claimLeafOutput
                             swapVC.thisSwap!.refundLeafOutput = refundLeafOutput
                             swapVC.thisSwap!.refundPublicKey = refundPublicKey
-                            swapVC.thisSwap!.sentLightningPaymentID = randomPreimage.hexEncodedString()
                             self.saveSwapDetailsToFile(swapID: swapID, swapDictionary: swapVC.thisSwap!.toDictionary())
                             
                             // Store transaction details in cache.
@@ -625,26 +624,6 @@ class SwapManager: NSObject {
         
         print("Updated swap file with lockup transaction for ID: \(swapID)")
             
-    }
-    
-    static func updateSwapFileWithFees(swapID: String, totalFees: Int, userAmount: Int, direction: Int) {
-        
-        // Load existing swap details
-        guard let existingSwapDetails = loadSwapDetailsFromFile(swapID: swapID) else {
-            print("Could not load existing swap details for ID: \(swapID)")
-            return
-        }
-        
-        // Create a mutable copy and add the fees information
-        let updatedSwapDetails = existingSwapDetails.mutableCopy() as! NSMutableDictionary
-        updatedSwapDetails.setValue(totalFees, forKey: "totalFees")
-        updatedSwapDetails.setValue(userAmount, forKey: "userAmount")
-        updatedSwapDetails.setValue(direction, forKey: "direction")
-        
-        // Save the updated swap details back to file
-        self.saveSwapDetailsToFile(swapID: swapID, swapDictionary: updatedSwapDetails)
-        
-        print("Updated swap file with fees for ID: \(swapID)")
     }
     
     static func addOnchainTransactionToUI(transactionId:String, swapVC:SwapStatusViewController) {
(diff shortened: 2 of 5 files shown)


RubenWaterman (commented):
Reviewed this, rebased it onto current `android`, and pushed a second commit (`040cc49`) acting on what the refactor exposed. Findings are recorded as they stood against `1f22ff9`, with what changed noted under each.

## Conflicts

Ninth PR forked from `f2e17d3` — and the first GitHub reported clean, since it touches only `Swap.swift` in regions #52 didn't. I rebased anyway (`--onto origin/android f2e17d3`) and **built it**: #52 landed changes to this exact file shortly before, and on #51 a clean textual merge still produced a branch that didn't compile. Zero conflicts here, builds clean.

## Overview

Collapses `toDictionary` and `toSwap` from ~142 lines of near-identical per-field blocks to one line per field, grouped to mirror the property declarations. Two claims underpin it:

1. `NSMutableDictionary.setValue` removes the key when handed `nil`, so `if x != nil { setValue(x!) }` is redundant.
2. Every optional property starts `nil` on a fresh `Swap`, so `if let x = d[k] as? T { swap.x = x }` is just `swap.x = d[k] as? T`.

## Both claims check out

**Claim 1 is documented behaviour** specific to `NSMutableDictionary`: `setValue(_:forKey:)` invokes `removeObject(forKey:)` when the value is nil. The Swift signature is `setValue(_ value: Any?, forKey:)` — `Any?`, not `Any` — so passing a `String?`/`Int?` flattens correctly rather than boxing an `Optional` as a non-nil object. That distinction is what would have made this subtly wrong, and the code is on the right side of it.

**Claim 2 holds** because `toSwap()` constructs `Swap()` fresh on every call, so assigning `nil` from a failed cast is genuinely a no-op. The four non-optional properties (`dateID`, `swapDirection`, `satoshisAmount`, `isSuggested`) correctly keep their `if let` guards so a missing key leaves the default intact — and a comment says so.

Verified rather than assumed:

- **Key sets are identical.** Every key the old `toDictionary` wrote, the new one writes; same for `toSwap`. The one difference: the old code set `boltzOnchainAddress` **twice** — once after `boltzID`, once after `sentOnchainTransactionID`. Deduplicating it is a no-op.
- **Direction encoding is equivalent.** `SwapDirection` has exactly two cases (`Transaction.swift:60-63`), so `swapDirection == .onchainToLightning` is the exact inverse of the old `== .lightningToOnchain` test. A third case would have silently flipped it.
- **Wrong-typed values behave the same.** A key present with the wrong type failed the old `as?` and left the property nil; the new form assigns nil. Identical.
- `satoshisAmount` / `isSuggested` moving from a dictionary literal to `setValue` still bridge to `NSNumber` either way.

A genuine simplification with no behavioural drift.

## What the refactor exposed

Flattening the persisted fields into a list that mirrors the declarations made one omission obvious, and the same shape turned up three more times across the surrounding PRs. All four were assigned and never read; all four are removed in `040cc49`, each verified unread first.

### `Swap.sentLightningPaymentID` — deleted

Declared, assigned once at `SwapManager.swift:513`, never persisted, never read. The assignment sits on the line immediately before `saveSwapDetailsToFile(... .toDictionary())`, which reads like it was meant to be saved.

The stronger reason to delete rather than persist: it's set to `randomPreimage.hexEncodedString()` — **the same value already assigned to `Swap.preimage` seventeen lines earlier in the same function**. `preimage` is persisted and read. So persisting this one would have written the same secret twice, under a key whose name claims it holds a payment ID.

### `Transaction.note` — deleted

Populated from the note cache on every transaction built, and read nowhere. `TransactionViewController` reads notes straight from `CacheManager`, so the field was a stale-data trap for whoever wired it up to a cell first — notes can be deleted (#46), while this snapshot is taken at list-build time.

### `WalletCache.balance` — deleted

Migrated across from the legacy `NSDictionary` cache with no accessor and no reader; the old stringly-typed cache carried the same unused key. Safe to drop from the `Codable` struct — decoding ignores keys it doesn't declare, so existing caches still load. This is the only one of the four with a data-shape consequence: a cache still holding `"balance"` silently loses it on the next write. Nothing read it.

### `SwapManager.updateSwapFileWithFees` — deleted

Writes `totalFees`, `userAmount` and `direction` into the swap file. **None of the three is read anywhere.**

This also resolves what I first wrote up as a separate nit — that the save/update round trip was lossy, because `toDictionary()` doesn't produce those keys and would drop them on the next save. Making `toDictionary()` preserve them would have been carefully preserving data nothing consumes. Removing the writer makes the round trip total instead.

**Scope note:** this second commit reaches beyond `Swap.swift` into `Transaction.swift`, `WalletCache.swift`, `SwapManager.swift` and `SwapViewController.swift`. It's deliberately a separate commit, so it can be dropped or split out if you'd rather keep this PR to one file. Net −31 lines on top of the refactor.

## Style

Good. The grouping comments (`// Boltz reply`, `// Fees`, `// Onchain to Lightning`, `// Lightning to Onchain`) mirror the property declarations exactly, so a missing field is now visible by inspection — which is how the first dead field above was found. The two explanatory comments earn their place: they state the non-obvious preconditions (`setValue` nil-removal, fresh-object defaults) that make the collapse safe.

## Testing

No tests, and — as with #49, #50 and #52 — this is trivially testable: `toDictionary()` / `toSwap()` is a pure round trip on a value object with no LDK, BDK, or network dependency. One test that builds a fully-populated `Swap`, round-trips it, and asserts field equality would pin the equivalence checked by hand here, and would immediately catch a property added to the class but forgotten in the conversions — exactly the failure mode that produced the dead fields above.

Compile-verified only. Manual pass:

- start an onchain→lightning swap, background the app, reopen → the swap resumes with its details intact
- same for lightning→onchain
- a swap file written by the **old** build must still load under the new code. The key sets are identical so it should, but it's the one thing worth confirming once by hand.

## Verdict

**Approve.** The cleanest PR in this series: a real 57-line reduction, both underlying claims are correct, and the field-by-field behaviour is preserved. The follow-up commit removes four write-only fields the refactor made visible — including one that was quietly storing a preimage under a name claiming it was a payment ID.

PR #54, opened 2026-07-22
Minor LoadWalletData refactoring

tommelters:
LoadWalletData.swift carried three extensions that had nothing to do with loading wallet data: PaymentKind (transactionID, isOnchain), a UILabel font helper, and the payment-status predicates used to decide which payments reach the table. Move them to Extensions/, where the rest of the codebase looks for that kind of thing, and name the status checks — hasSucceeded, isPendingOutbound, isUnconfirmedOnchainInbound — so the filter reads as three conditions rather than one long boolean.

createTransaction took a coreVC it never used, on both the PaymentDetails and the BittrTransaction overload. Dropped, along with the argument at every call site.

The rest is local tidying with no behaviour change: `for ... where` in place of a loop wrapping a single if, ternaries for two-branch assignments, and flattened nesting in isValidInvoice and bolt12Offer.

### DIFF
--- a/ios/bittr.xcodeproj/project.pbxproj
+++ b/ios/bittr.xcodeproj/project.pbxproj
@@ -36,6 +36,9 @@
 		05117AF72EAA754800A5E4E6 /* Demo Data.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05117AF62EAA754500A5E4E6 /* Demo Data.swift */; };
 		05117AF92EAB74A100A5E4E6 /* Image.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05117AF82EAB749F00A5E4E6 /* Image.swift */; };
 		05117B9E2EDEC6AC00A5E4E6 /* Log.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05117B9D2EDEC6AA00A5E4E6 /* Log.swift */; };
+		0514BC5330109C3F00BC05A2 /* PaymentDetails.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0514BC5230109C3B00BC05A2 /* PaymentDetails.swift */; };
+		0514BC553010A17D00BC05A2 /* PaymentKind.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0514BC543010A17B00BC05A2 /* PaymentKind.swift */; };
+		0514BC573010A1AF00BC05A2 /* UILabel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0514BC563010A1AE00BC05A2 /* UILabel.swift */; };
 		051AB88B2F5E816B00DFC39F /* ConfirmSendViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 051AB88A2F5E816B00DFC39F /* ConfirmSendViewController.swift */; };
 		051AB88F2F5E8FB400DFC39F /* ConfirmLanguage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 051AB88E2F5E8FB100DFC39F /* ConfirmLanguage.swift */; };
 		051BE0882B641D41009CED5B /* Reachability.swift in Sources */ = {isa = PBXBuildFile; fileRef = 051BE0872B641D41009CED5B /* Reachability.swift */; };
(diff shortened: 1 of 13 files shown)


RubenWaterman (commented):
Reviewed this and pushed a second commit (`dabdcd5`) restoring two rationales the move dropped. Findings are recorded as they stood against `526587e`, with what changed noted under each.

## Base — actually current

The one PR in this series that isn't stale. Merge-base is `855899d`, today's `android` HEAD, with #53 already in it. No rebase needed, the diff stat is real (no phantom inflation from a stale base), and it **builds clean** against current `android`. Worth stating plainly since every earlier PR in this batch reported clean and still needed a rebase — this one genuinely doesn't.

## Overview

Three misplaced extensions leave `LoadWalletData.swift` for `Extensions/`: `PaymentKind` (`transactionID`, `isOnchain`), the `UILabel.adjustedFont` helper, and the payment-status predicates. The three-way payment filter is named into `hasSucceeded()` / `isPendingOutbound()` / `isUnconfirmedOnchainInbound()`. `createTransaction`'s unused `coreVC` parameter is dropped on both overloads. Plus local tidying — `for … where`, ternaries, flattened guards.

## Correctness — behaviour is preserved

Checked the parts that could drift:

- **The named predicates are exact extractions.** `hasSucceeded || isPendingOutbound || isUnconfirmedOnchainInbound` reassembles the original one-line boolean term for term — same operator precedence, same `?? 0` guards. This is the change most worth getting right, and it's right.
- **`coreVC` really was unused.** Zero references in either `createTransaction` body, and all 9 call sites are updated — none left passing it. Dead parameter, cleanly removed.
- **The dedup simplification is equivalent.** Collapsing `if a { if b { … } }` to `if a, b, c { … }`, and the inner loop to `for … where`, preserves the guard: `transactionID != nil` still short-circuits before the force-unwrap on the next line.
- **`CacheManager` `for (_, transactionData)`** — the discarded key was genuinely unused; correct `_` replacement.

## Findings

### Low — `bolt12Offer()` silently dropped a diagnostic

The old `catch` logged `Log.info("Could not generate BOLT12 offer.")`; the refactor removed it. Behaviour to the caller was unchanged, but this path is reached from `isValidInvoice()` while parsing pasted/scanned input, so a malformed `lno…` string that used to leave a breadcrumb would fail silently — an unannounced removal in a PR describing itself as behaviour-preserving tidying.

**Restored** in `dabdcd5`, with a line noting where it's reached from.

### Low — a rationale comment was thinned to a restatement

`LoadWalletData.swift:519`, before:

> Skip the payout check when a wipe / PIN reset is in progress: the node is about to be torn down, so signing a message against it would race the teardown (force-unwrap of a nil ldkNode).

after:

> Skip payout check when a wipe / PIN reset is in progress.

The *what* survived; the *why* — that this specifically avoids a nil-`ldkNode` crash during teardown — was gone. That's exactly the reason someone needs before they "simplify" the guard away.

**Restored** in `dabdcd5`. This is the fourth time in this series a refactor dropped a comment explaining a non-obvious safety invariant (#45's force-close notes, #47's watchdog explanation, #50's fee comment). Worth raising as a habit rather than catching per-PR: a mechanical refactor should carry rationale comments across untouched.

### Style — the extracted predicates keep the `if x { return true } else { return false }` shape

`hasSucceeded()` and its siblings were moved verbatim, so they carry the old `if condition { return true } else { return false }` pattern into their new home. Since the PR is explicitly naming and clarifying these, `return self.status == .succeeded` would be the natural finish. **Left as-is** — consistency with the source they were lifted from is a fair reason, and reshaping the author's just-created file beyond the actual regressions above is overreach. Flagging as a your-call cleanup.

### Nits

- **`PaymentDetails.swift` and `PaymentKind.swift` both `import UIKit`** — neither needs it; `LDKNode` + `Foundation` suffices. Same latent habit noted on #49/#50/#52. `UILabel.swift` genuinely does need `UIKit`.
- **Two extensions named `extension PaymentDetails`** now exist — the new predicates file and the existing `createTransaction` in `Transaction.swift`. Both valid; just worth knowing the type's surface is split across two files by two different axes (parsing vs. status).

## Testing

No tests — but the extracted pieces are now trivially reachable, which is the point of moving them. `PaymentKind.transactionID` across all six variants, `isOnchain`, and the three status predicates are pure functions on enum/struct values with no LDK-node or network dependency once a `PaymentDetails` is in hand. The predicates decide which payments appear in the transaction list, so a table covering succeeded / pending-outbound-with-amount / pending-outbound-zero / unconfirmed-onchain-inbound would lock the filter this PR just isolated.

Compile-verified only. Manual pass, given it touches the transaction-list filter and every `createTransaction` call site:

- transaction list still shows the same set after a sync (the predicate extraction)
- an incoming lightning payment still opens its `TransactionVC` (the `coreVC`-drop call sites in `HandlePaymentNotification`)
- paste a BOLT12 offer and a bolt11 invoice into Send — both still parse (`isValidInvoice` / `bolt12Offer`)

## Verdict

**Approve.** Well-scoped, current, and behaviour-preserving where it counts — I verified the payment filter and the parameter drop, the two places a "just moving things" PR could quietly change what ships. The two dropped rationales are restored on the branch; everything else is cosmetic or a your-call cleanup.

PR #55, opened 2026-07-30
Retry transient errSecMissingEntitlement on Keychain reads

tommelters:
This is a fix for **this Sentry error**: https://bittr.sentry.io/issues/7636076407/?project=4507055778758656

This error was thrown in **CacheManager row 820**.

**Reason for Sentry error**
SecItemCopyMatching can spuriously return errSecMissingEntitlement (-34018) when queried very early in the launch cycle — a cold launch or an OS background launch (the app declares fetch/remote-notification) — before securityd has finished wiring up the app's keychain-access-group entitlement. This surfaced as SecureStoreError.unexpectedStatus(-34018) from walletSecretsPresence() at CoreViewController launch.

**The fix: 3 quick retries before throwing an error**
Add a bounded retry inside SecureStore.getData for that status only (3 retries, ~210 ms worst case), so a transient failure isn't reported as a real one. Success, not-found, and every other status (including a persistent -34018 after retries) behave as before.

### DIFF
--- a/ios/bittr/Helpers/SecureStore.swift
+++ b/ios/bittr/Helpers/SecureStore.swift
@@ -122,6 +122,11 @@ enum SecureStore {
         }
     }
 
+    /// Retry schedule (seconds) for a transient read failure — see below.
+    /// Three retries with escalating backoff, ~210 ms of blocking in the worst
+    /// case, which only ever elapses when every attempt fails.
+    private static let readRetryDelays: [TimeInterval] = [0.03, 0.06, 0.12]
+
     static func getData(account: String) throws -> Data? {
         let query: [String: Any] = [
             kSecClass as String:                     kSecClassGenericPassword,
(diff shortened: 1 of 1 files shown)


RubenWaterman (approved):
## Overview

Fixes a Sentry-reported crash where `SecItemCopyMatching` spuriously returns `errSecMissingEntitlement` (-34018) when the Keychain is read very early in the launch cycle, before `securityd` has wired up the app's keychain-access-group entitlement. The fix wraps `SecureStore.getData` in a bounded retry loop: on -34018 only, it retries up to 3 times with escalating backoff (`[0.03, 0.06, 0.12]`s, ~210 ms worst case). Success, not-found, and every other status — including a persistent -34018 after all retries — behave exactly as before.

## Code Quality & Correctness

Clean and well-documented — the comments explain *why* (the securityd race) rather than just *what*, matching the file's existing style.

- **Retry logic is correct.** The `where attempt < readRetryDelays.count` guard means attempts 0/1/2 sleep-and-retry, and attempt 3 falls through to `default` and throws. Exactly 3 retries, 210 ms total.
- **Scope is right.** Retrying only reads (not `setData`) is defensible — the crashing path is a read (`walletSecretsPresence`), and reads are idempotent so retrying is always safe.
- **Only -34018 is retried.** All other statuses throw immediately, so a missing item or a real error isn't masked or delayed.

## Main consideration — main-thread blocking

`getData` uses `Thread.sleep`, and the crashing caller runs on the main thread:

`CoreViewController.viewDidLoad` → `checkWalletAvailability()` → `CacheManager.walletSecretsPresence()`, which reads **two** secrets sequentially (`mnemonic` then `pin`).

- In the **transient-then-recovers** case (what this PR targets), you pay one or two short sleeps — 30–90 ms. Fine.
- In the **persistent -34018** worst case, each read exhausts all retries: ~210 ms × 2 = **~420 ms of main-thread blocking at launch** before falling through to `.unavailable`. That's during the launch animation, so largely invisible, and well under the watchdog threshold — but a real UI stall on a genuinely broken device.

Acceptable given the constraints; no change strictly required. Optionally, the backoff could be capped so both reads share a single budget.

## Minor notes

- **Base branch is `android`, not `master`** — worth confirming that's intentional so the fix reaches the release branch.
- **No tests** — understandable; the securityd race isn't practically unit-testable and the retry constants are self-documenting. Not a blocker.

## Verdict

Correct, minimal, well-scoped fix that addresses the reported error without changing behavior for any non-transient status.

PR #56, opened 2026-07-30
Fix and rotate the LDK Node log file

tommelters:
deleteLDKNodeLogLatestFile pointed at logs/ldk_node_latest.log, a path LDK Node never writes — the real file is ldk_node.log in the Documents dir — so the pre-start removal silently failed every launch and the log grew without bound.

Point it at the real file, and rotate rather than hard-delete: the previous session is preserved as ldk_node_previous.log so a crash or channel close can still be investigated after a relaunch.

### DIFF
--- a/ios/bittr/BitcoinManager.swift
+++ b/ios/bittr/BitcoinManager.swift
@@ -116,8 +116,8 @@ class BitcoinManager {
     
     func didStartLDK() -> Bool {
         
-        // Delete previous LDK Node log.
-        try? FileManager.deleteLDKNodeLogLatestFile()
+        // Rotate the previous LDK Node log.
+        try? FileManager.rotateLDKNodeLog()
         
         // Congifure LDK Node settings.
         let correctListeningAddresses = EnvironmentConfig.isDevelopment ? ["0.0.0.0:19735"] : ["0.0.0.0:9735"]
--- a/ios/bittr/Extensions/FileManager.swift
+++ b/ios/bittr/Extensions/FileManager.swift
@@ -9,8 +9,11 @@ import Foundation
 
 extension FileManager {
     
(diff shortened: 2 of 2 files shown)


RubenWaterman (approved):
## Overview

Fixes `deleteLDKNodeLogLatestFile()`, which pointed at `logs/ldk_node_latest.log` — a path LDK Node never writes — so the pre-start `removeItem` threw every launch (silently, since the caller uses `try?`) and the log grew unbounded. The fix points at the correct file and rotates instead of deletes: the current log becomes `ldk_node_previous.log`, preserving the prior session for debugging.

**Path claim verified** against the pinned dependency (ldk-node v0.7.0): `config.rs` defines `DEFAULT_LOG_FILENAME = "ldk_node.log"`, and `builder.rs` builds the default logger path as `<storage_dir_path>/ldk_node.log`. The app sets `storageDirPath` to `getDocumentsDirectory()` and leaves `setFilesystemLogger` commented out, so the real file is exactly `<Documents>/ldk_node.log`. The old path was dead.

## Correctness

- Rotation logic is sound: `guard fileExists else return` (clean first-launch), `try? removeItem(previous)` (tolerates not-found), `try moveItem(current → previous)`.
- Ordering is right: rotation runs at the top of `didStartLDK()`, before `build()`/`start()` open the new session's log handle, and the prior session's process is gone at relaunch — no open-handle contention.
- Resolves the unbounded growth that motivated the PR.

## Follow-up commit

Pushed `rotateLDKNodeLog` rename (+ call site) and wrapped the doc comment — the old name no longer described the behavior (rotate, not delete; no "latest" file).

## Out of scope (separate follow-up, not blocking)

- Within a single long-running session the active log still grows unbounded — rotation only happens at start.
- LDK logs can contain channel/routing detail and `Documents/` is backed up by default; retained footprint now roughly doubles (two sessions vs. intended zero). Consider excluding from backup / lowering default log level if that matters.

## Verdict

Correct, well-targeted fix, verified against the dependency source. Approving.

PR #57, opened 2026-07-30
Present the QuestionVC from the top-most view controller

tommelters:
launchQuestion fired a CoreToQuestion segue from CoreViewController, which sits at the root beneath the Home / Move / Swap modal chain. Presenting from a controller that is buried under other modals is unreliable (and impossible mid-transition), so the screen silently failed to appear over deeper modals — e.g. the channel-closed alert during a swap.

Instead, instantiate QuestionViewController by storyboard id and present it from the top-most presented controller, whose view is always frontmost and can always present. This makes every launchQuestion call site robust regardless of what is on screen.

Remove the now-dead path: the CoreToQuestion segue, its prepare(for:) branch in Navigation, and the tappedQuestion/Answer/Type properties.

### DIFF
--- a/ios/bittr/Base.lproj/Main.storyboard
+++ b/ios/bittr/Base.lproj/Main.storyboard
@@ -829,7 +829,6 @@
                         <outlet property="walletView" destination="LZ5-LN-ll3" id="YCy-ng-lAQ"/>
                         <outlet property="yearLabel" destination="IaO-Gf-3C2" id="OrV-1q-QoO"/>
                         <outlet property="yearView" destination="gSx-Va-0OK" id="zr4-cO-epd"/>
-                        <segue destination="b12-fa-pJV" kind="show" identifier="CoreToQuestion" id="KXI-1X-aAi"/>
                         <segue destination="4ca-QY-6dF" kind="show" identifier="CoreToArticle" id="o32-HH-NM5"/>
                         <segue destination="SiB-n5-gpJ" kind="show" identifier="CoreToSwap" id="ahR-EC-NqA"/>
                     </connections>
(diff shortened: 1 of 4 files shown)


RubenWaterman (approved):
## Overview

Replaces the segue-based launch of `QuestionViewController` with programmatic presentation from the top-most presented controller. The old `CoreToQuestion` "show" segue fired from `CoreViewController` — the root, buried under the Home/Move/Swap modal chain — so the question screen silently failed to appear over deeper modals (the motivating case: a channel-closed prompt during a swap). Now it instantiates the VC by storyboard id and presents from whatever is frontmost. Also removes the dead segue, its `prepare(for:)` branch, and the `tappedQuestion/Answer/Type` staging properties.

## Code Quality & Correctness

- **Diagnosis is right, fix is the standard remedy.** Presenting from a controller buried under modals is genuinely unreliable; walking to the top-most presented VC is the correct approach.
- **Clean removal** — no dangling references to `tappedQuestion/Answer/Type` or `CoreToQuestion` anywhere. `storyboardIdentifier="Question"` matches `instantiateViewController(withIdentifier: "Question")`.
- **No retain cycle** — `questionVC.coreVC = self` is a strong ref to the root `CoreViewController`, which outlives the modal; `CoreViewController` holds no reference back, so `questionVC` deallocs normally on dismiss.
- **`topMostViewController()` is correct** for this layout — it follows the `.presentedViewController` chain from the key window's root, which is exactly the modal stack in play.
- Style matches the codebase (extension on `CoreViewController`, `guard let`, `Log.info`).

## Alert-on-top — researched, not a concern

Confirmed the app has exactly two real `UIAlertController`s (`SendLNURL.swift:250` LNURL-withdraw, `ReceiveViewController.swift:600` action sheet). Everything else called an "alert" (the app's own `showAlert` / AlertManager, the `alertCard`/`alertStack` views) is a custom subview overlay, which a modal presents over cleanly.

Critically, the swap / channel-closed path — the exact bug this fixes — uses **no** `UIAlertController` (none in `Swaps/` or `Notifications/`), so `topMostViewController()` returns a normal VC there. The only residual edge is a push arriving while one of those two transient, non-swap alerts is open; even then it's not a regression (the old buried-segue failed there too), and every attempted guard is worse than the benign default. Leave as-is.

## Other notes (non-blocking)

- **Presentation style** is now the `present` default (`.pageSheet`, swipe-dismissable). Consistent with the VC's existing down-button dismiss; verify it reads correctly if the old "show" segue rendered full-screen.
- **Mid-transition** presents can still silently no-op (same failure class, tap/notification-driven — acceptable).
- **Repeated calls stack** a second `QuestionViewController` over the first. Minor.
- **Main-thread dependency** (`UIApplication.shared`, `present`) — same as the old `performSegue`, no regression.
- The per-call `Log.info("presenting from …")` reads like debug instrumentation; fine to trim once confirmed in the wild.

## Test coverage & Security

- No tests — UIKit navigation, awkward to unit-test; acceptable. Manual check: trigger a question over the deepest swap modal (the repro) and confirm it appears and dismisses.
- No security surface.

## Verdict

Solid, well-targeted fix that removes a real reliability bug and leaves no dead code behind. The alert-on-top question I raised was researched and does not affect the targeted scenario. Approving.

PR #58, opened 2026-07-30
Fix Academy locked-lesson blur render + live dark-mode repaint

tommelters:
Fixes the Academy locked-lesson blur render, and — while in the same files — a pre-existing dark-mode repaint gap.

## 1. Blur render (the original fix)

`f2e17d3`'s "build each locked-lesson blur once per attachment" (`isConfigured`) froze the blur at the size it had in `didMoveToSuperview`, which runs during `cellForItemAt` before the cell is laid out. AutoLayout then stretched that premature, low-resolution render up to the tile size, producing oversized, misaligned, blocky blur patches.

Rather than reverting to rebuild-on-every-post (which renders correctly but reintroduces the main-thread hang the `isConfigured` guard was added to prevent — rebuilding a `UIVisualEffectView` blur is an expensive filter regeneration, and `"setupblur"` fans out to every attached instance per post), the rebuild is now **keyed on size**:

- The blur is built in `layoutSubviews`, once the real post-layout tile size is known — correct render.
- Repeated same-size `"setupblur"` posts (and re-render storms, e.g. a wallet reset re-rendering the academy) become cheap no-ops — keeps the watchdog-hang protection.
- Foreground still forces one rebuild (`invalidateBlur` clears the cached size), since backgrounding can drop the paused effect.

The valuable half of the earlier work — recreating the animator instead of accumulating `addAnimations` blocks — is preserved.

## 2. Live dark-mode repaint (pre-existing gap, fixed here)

The Academy screen and its cells only set colors in `viewDidLoad` / `awakeFromNib` and subscribed to no repaint notification, so toggling dark mode in Settings left the Academy showing stale colors until cells were re-created by scrolling. Every other list screen (Home, Settings, Pin, History cells, …) already observes `"changecolors"` and repaints.

`AcademyViewController`, `LevelTableViewCell`, and `LessonCollectionViewCell` now observe `"changecolors"` and re-run `changeColors`, matching the app-wide pattern — so returning to the Academy after a dark-mode change shows the correct colors immediately. The blur tiles stay light-pinned by design; only the card/text/tint colors flip.

### DIFF
--- a/ios/bittr/Academy/Academy/AcademyVC/AcademyViewController.swift
+++ b/ios/bittr/Academy/Academy/AcademyVC/AcademyViewController.swift
@@ -31,6 +31,9 @@ class AcademyViewController: UIViewController, UITableViewDelegate, UITableViewD
         self.academyTableView.rowHeight = UITableView.automaticDimension
         self.academyTableView.estimatedRowHeight = 300
         
+        // Repaint on dark-mode change, like the app's other screens.
+        NotificationCenter.default.addObserver(self, selector: #selector(changeColors), name: NSNotification.Name(rawValue: "changecolors"), object: nil)
+
         // Get Academy levels.
         self.getLevels()
         self.changeColors()
@@ -132,7 +135,7 @@ class AcademyViewController: UIViewController, UITableViewDelegate, UITableViewD
         }
     }
     
-    func changeColors() {
+    @objc func changeColors() {
         self.view.backgroundColor = Colors.getColor("yelloworblue3")
         self.headerLabel.textColor = Colors.getColor("blackorwhite")
     }
--- a/ios/bittr/Academy/Academy/AcademyVC/LessonCollectionViewCell.swift
+++ b/ios/bittr/Academy/Academy/AcademyVC/LessonCollectionViewCell.swift
@@ -35,7 +35,10 @@ class LessonCollectionViewCell: UICollectionViewCell {
         
         // Resistance priority.
         self.lessonTitle.setContentCompressionResistancePriority(.required, for: .vertical)
-        
+
+        // Repaint on dark-mode change, like the app's other list cells.
+        NotificationCenter.default.addObserver(self, selector: #selector(changeColors), name: NSNotification.Name(rawValue: "changecolors"), object: nil)
+
         // Color management.
         self.changeColors()
     }
@@ -64,8 +67,8 @@ class LessonCollectionViewCell: UICollectionViewCell {
         }
     }
     
-    func changeColors() {
-        
+    @objc func changeColors() {
+
         self.lessonTitle.textColor = Colors.getColor("blackorwhite")
         
         if CacheManager.darkModeIsOn() {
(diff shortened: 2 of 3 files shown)


RubenWaterman (commented):
## Review summary

Note on authorship: commits 2 and 3 below were written during this review session, so this is a summary/assessment rather than an independent approval.

### Commit 1 — Revert the size-freezing `isConfigured` guard (`a8990df`)
Correctly diagnoses the render bug: `isConfigured` built the blur once in `didMoveToSuperview` (pre-layout), and AutoLayout stretched that premature render into oversized/blocky patches. Removing it fixes the render but, on its own, gives back the main-thread hang the guard was added to prevent.

### Commit 2 — Key the rebuild on size (`b416a56`)
Resolves that trade-off. The blur is built in `layoutSubviews` once the real tile size is known (correct render), and `setupBlur` early-returns when `bounds.size` is unchanged, so the `willDisplay` / re-render storm (e.g. a wallet reset) collapses to cheap no-ops — retaining the watchdog-hang protection. Foreground still forces one rebuild via `invalidateBlur`, covering the dropped-effect case. The animator-recreation fix is preserved.

### Commit 3 — Live dark-mode repaint (`be21daa`)
Pre-existing gap: the Academy subscribed to no repaint notification, so a dark-mode change left it stale until cells were re-created by scrolling. Now `AcademyViewController`, `LevelTableViewCell`, and `LessonCollectionViewCell` observe `"changecolors"` and re-run `changeColors`, matching the pattern already used by Home / Settings / History.

## Correctness / side-effect checks
- Observers are added once (`awakeFromNib` per cell instance, `viewDidLoad` per VC) — no stacking on reuse.
- The two `removeObserver(self)` calls in the file belong to `BlurEffect`, a separate object — they don't clear the cell's `"changecolors"` observer.
- `"changecolors"` only sets `blurView.backgroundColor`; it doesn't rebuild the blur or change its bounds, so it can't trigger the size-keyed rebuild. No blur/colour interaction.
- The VC handler calls `changeColors()` only (not `reloadData()`), so it doesn't cascade into `setupblur`.
- Selector-based observers are auto-released on dealloc (iOS 9+), consistent with the rest of the app. `@objc` is additive, no name collision.

## Testing
- Blur render + hang: on a real device, trigger a wallet-reset academy re-render — blur crisp and tile-aligned, no main-thread stall.
- Dark-mode repaint: open Academy → toggle dark mode in Settings → return to Academy; colors are correct immediately (no scroll needed).

Base branch is `android` (consistent with the related PRs).

RubenWaterman:
@tommelters — I pushed two commits on top of your revert, and updated the PR title/description to match. Could you review when you get a chance?

- `b416a56` — instead of rebuilding the blur on every `"setupblur"` post, key the rebuild on `bounds.size` (built in `layoutSubviews` at the real tile size). Keeps your correct render but restores the re-render/hang protection the old `isConfigured` guard gave.
- `be21daa` — wire the Academy (`AcademyViewController`, `LevelTableViewCell`, `LessonCollectionViewCell`) into the `"changecolors"` notification so it repaints live on a dark-mode change, matching Home/Settings/History. Pre-existing gap, fixed here since we were in the same files.

Worth a device check before merge: a wallet-reset academy re-render (blur crisp, no main-thread stall) and returning to the Academy after toggling dark mode.

tommelters:
@RubenWaterman This works nicely🙏

PR #59, opened 2026-07-31
Show pending lightning balances from channel closures

tommelters:
**The problem**
When a channel closes, its funds leave `listChannels()` immediately but take a while to arrive on-chain. The Instant balance dropped to 0 for that whole window — up to about a day after a force close, since the `to_local` output sits behind the CSV delay.

**The solution**
BittrWallet now carries **pendingBalancesFromChannelClosures**, accumulated in loadWalletData(). 
The **HomeVC** and **MoveVC** both include it. 
The **MoveVC** greys the Instant label while any of it is pending and explains it in an alert on tap.

**What counts as pending**
Only funds a **force close** has put out of reach — locked in a script the on-chain wallet doesn't own, so nothing else counts them until the sweeper brings them back. 
A **cooperative close** pays our own shutdown script, so it lands in the  on-chain balance as unconfirmed within seconds.

**Improvements to lightSync**
The light sync only refreshed when the on-chain balance or the payment count changed. After a force close neither does for the whole CSV delay, so the 30s timer never reached `loadWalletData()` and both screens kept showing a stale balance for about a day. It now compares the pending total as well.

### DIFF
--- a/ios/bittr/AlertManager.swift
+++ b/ios/bittr/AlertManager.swift
@@ -196,6 +196,10 @@ extension UIViewController {
             yellowCard.addConstraints([messageLabelLeft, messageLabelRight, messageLabelTop])
             messageLabel.addConstraints([messageLabelHeight])
             
+            // Check whether to stack buttons vertically or horizontally.
+            let titlesFitSideBySide = buttons.allSatisfy { eachButton in eachButton.count < 16 }
+            let stackButtonsVertically = buttons.count > 2 || (buttons.count == 2 && !titlesFitSideBySide)
+            
             // Buttons stack
             let buttonsStack = UIView()
             buttonsStack.translatesAutoresizingMaskIntoConstraints = false
@@ -203,10 +207,10 @@ extension UIViewController {
             buttonsStack.backgroundColor = .clear
             yellowCard.addSubview(buttonsStack)
             let buttonsStackHeight:NSLayoutConstraint = {
-                if buttons.count < 3 {
-                    return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50)
-                } else {
+                if stackButtonsVertically {
                     return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50*CGFloat(buttons.count))
+                } else {
+                    return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50)
                 }
             }()
             let buttonsStackLeft = NSLayoutConstraint(item: buttonsStack, attribute: .leading, relatedBy: .equal, toItem: yellowCard, attribute: .leading, multiplier: 1, constant: 15)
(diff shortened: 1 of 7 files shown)


RubenWaterman (commented):
## Review

Note on authorship: commit `e094438` (the `.coopClose` change below) was made during this review session, so this is a summary/assessment rather than an independent approval.

Fixes the Instant balance dropping to 0 after a channel closes, for the window where funds have left `listChannels()` but haven't landed on-chain (up to ~1 day behind a force-close CSV delay). Adds `BittrWallet.pendingBalancesFromChannelClosures`, computed from LDK `BalanceDetails`, folds it into the Home/Move totals, greys the Move Instant label while pending, and makes light sync compare the pending total so it refreshes during the CSV window.

### Correctness — verified against the ldk-node bindings
- `BalanceSource = {holderForceClosed, counterpartyForceClosed, coopClose, htlc}`. The code counts the three force/HTLC sources and excludes `coopClose` — correct, since a coop close pays your shutdown script and already lands in `totalOnchainBalanceSats`.
- Double-count guard is sound: `satoshisLightning` comes from open `listChannels()` only, and `forceClosedSatoshis` skips any `channelId` still in the open set, so a channel mid-close isn't counted twice.
- The `broadcastAwaitingConfirmation` / `awaitingThresholdConfirmations` exclusions correctly avoid double-counting with the on-chain wallet (those enter `totalOnchainBalanceSats` once broadcast).

### Applied during review
- **`.coopClose` now matched explicitly** (commit `e094438`) instead of via a bare catch-all. The switch is now exhaustive over `BalanceSource`, so a future LDK case is a compile error to classify rather than a silent exclusion of real funds.

### Remaining considerations (non-blocking)
- **`maybe*` HTLCs (`maybeTimeoutClaimableHtlc` / `maybePreimageClaimableHtlc`)** are the balances LDK flags as *uncertain*. Counting them is optimistic — if one resolves against us the balance later drops. Fine to keep, but worth a code comment (or excluding the pair if you'd rather under-promise).
- **Brief dip at broadcast:** the instant a sweep tx is broadcast it leaves `pending` but only enters `totalOnchainBalanceSats` after the next BDK sync (up to ~30s) — a short version of the gap this PR fixes.
- **Naming:** `BittrWallet.pendingBalancesFromChannelClosures` (Int) collides with LDK's `BalanceDetails.pendingBalancesFromChannelClosures` ([PendingSweepBalance]); `pendingClosureSatoshis` would read clearer.
- **Live refresh:** confirm MoveVC re-runs `updateLabels()` + `changeColors()` when balances update via light sync while the Move screen is on screen, so the grey/amount reflect a closure that happens while the user is looking at it.
- Unrelated: the `satoshisOnchainSpendable` comment was trimmed of its nil-vs-zero rationale — worth keeping.

### Testing
- **No Maestro coverage of this behavior.** The existing close-channel flows (`remove_wallet.yaml`, `wrong_pin_with_channel.yaml`) all do *cooperative* closes — precisely the case this PR excludes — and none assert the pending Instant balance, the greyed label, or the "Connection closed" / "View active connection" alert.
- **TODO: add a Maestro test for the force-close pending-balance flow.** It needs the regtest harness to *force*-close a channel (not coop) and leave outputs behind the CSV delay, then assert the greyed Instant balance and the new alert. Adding `accessibilityIdentifier`s to the new alert buttons first would make that flow robust (it currently keys off visible text only).
- In the meantime, unit tests on the pure functions (`forceClosedSatoshis`, `unbroadcastSatoshis`, `pendingClosureSatoshis`) with representative force/coop/open-channel/swept arrays would lock in the money classification cheaply and guard the exhaustiveness invariant above.

### Verdict
Solid, correctly-reasoned fix for a real balance-display bug, verified against the LDK bindings, with the money-path exhaustiveness now enforced at compile time. Remaining items are minor. Base branch is `android`, consistent with the related PRs.

PR #60, opened 2026-07-31
Distinguish onchain payout from a closed lightning connection in TransactionVC

tommelters:
**The problem**
Funds returning from a closed channel arrived as a plain incoming transaction with nothing to explain them — most starkly after a force close, where they show up a day after the connection disappeared from the app. 

**Solution**
The **TransactionVC** now explains, in the **Description**, that these funds result from a channel closure.

**How does that work?**
Recognising which transaction that is works differently for each kind of close.

**Force close.** LDK names it: the sweep carries a `latestSpendingTxid` in `listBalances()`. But only while the sweep is pending — by the time the transaction reaches the history there's nothing left to recognise it by — so the IDs are cached as they're seen.

**Cooperative close.** LDK never names it, because those funds pay our shutdown script directly and never pass through the sweeper. What identifies it is that it spends the channel's funding output, which only the BDK wallet holds. So the funding outpoint is cached while the channel is open, and once the channel is gone we look up the transaction spending it. The vout is part of the outpoint on purpose: the funding transaction's other output is our own change, and spending that is an ordinary payment.

**Onchain address management fix**
Address management derived the address pool with `peekAddress`, which returns the address but leaves the wallet's revealed index untouched. A light sync only asks the server about revealed addresses, so a payment to one of these was invisible to the BDK wallet until the next full scan — and that only runs at launch. They're revealed now, and persisted. Also a little past the pool, since closing a channel makes LDK reveal a fresh address for the shutdown script, one index beyond anything the pool reached.

### DIFF
--- a/ios/bittr/AlertManager.swift
+++ b/ios/bittr/AlertManager.swift
@@ -196,6 +196,10 @@ extension UIViewController {
             yellowCard.addConstraints([messageLabelLeft, messageLabelRight, messageLabelTop])
             messageLabel.addConstraints([messageLabelHeight])
             
+            // Check whether to stack buttons vertically or horizontally.
+            let titlesFitSideBySide = buttons.allSatisfy { eachButton in eachButton.count < 16 }
+            let stackButtonsVertically = buttons.count > 2 || (buttons.count == 2 && !titlesFitSideBySide)
+            
             // Buttons stack
             let buttonsStack = UIView()
             buttonsStack.translatesAutoresizingMaskIntoConstraints = false
@@ -203,10 +207,10 @@ extension UIViewController {
             buttonsStack.backgroundColor = .clear
             yellowCard.addSubview(buttonsStack)
             let buttonsStackHeight:NSLayoutConstraint = {
-                if buttons.count < 3 {
-                    return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50)
-                } else {
+                if stackButtonsVertically {
                     return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50*CGFloat(buttons.count))
+                } else {
+                    return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50)
                 }
             }()
             let buttonsStackLeft = NSLayoutConstraint(item: buttonsStack, attribute: .leading, relatedBy: .equal, toItem: yellowCard, attribute: .leading, multiplier: 1, constant: 15)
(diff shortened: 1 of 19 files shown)


RubenWaterman (commented):
## Review

Note on authorship: commit `573e177` (the two hardening changes below) was made during this review session, so this is a summary/assessment rather than an independent approval.

Labels the on-chain transaction returning funds from a closed lightning connection. Detection is split by close type: **force close** via the sweep's `latestSpendingTxid` from `listBalances()` (cached as seen), **cooperative close** via a background BDK scan for the tx spending the channel's cached funding outpoint. TransactionVC then shows an explanatory Description for those txids. Also reveals (not just peeks) on-chain addresses so payouts are visible to light sync.

### Correctness — verified
`Transaction.id` for on-chain payments is the txid (`Transaction.swift:79`), and the cached closure txids are on-chain txids (coop: `computeTxid()`, force: `latestSpendingTxid`), so `isChannelClosure` matching is sound — **provided the txid string formats agree** across BDK `computeTxid()`, LDK `latestSpendingTxid`, and `PaymentKind.transactionID` (all should be display byte-order hex). Worth one manual check, since the whole feature hinges on it.

The coop-only scan is correctly reasoned: a coop-close tx pays a BDK-owned shutdown script (so BDK sees it and its funding-outpoint input can be matched), while a force-close commitment tx has no BDK-owned output and is left to the sweep-txid path. The vout is correctly part of the outpoint match (the funding tx's other output is your change).

### Applied during review (commit 573e177)
- **`.coopClose` matched explicitly** in `forceClosedSatoshis` — the switch is now exhaustive over `BalanceSource`, so a future LDK case is a compile error to classify rather than a silent exclusion of real funds. (Same fix as on #59.)
- **Stale funding outpoint after a force close now cleared.** Previously the coop-close scan could never match a force close (commitment tx isn't BDK-visible), so the outpoint was never dropped and `storeChannelClosureTxIDIfFound` re-scanned every wallet tx on every sync indefinitely. Pending closure funds (> 0) only arise from a force close, so the outpoint is dropped there; a coop close leaves it at 0 and keeps the scan.

### Remaining considerations (non-blocking)
- **Rebase onto #59.** #60 re-adds the identical `pendingClosureSatoshis` / `forceClosedSatoshis` / `unbroadcastSatoshis` extensions #59 adds, so they'll conflict on merge; rebasing also reconciles the `.coopClose` change now present on both.
- **BDK `Wallet` thread-safety:** `storeChannelClosureTxIDIfFound` reads `bdkWallet.transactions()` on a background queue. It's called from sync completions, but confirm no concurrent sync/reveal/persist can touch the wallet at the same time (serialize on a queue, or accept the small race).
- **Address revealing widens the light-sync query set.** `getAddress(atIndex:)` reveals up to `atIndex` on every call — make sure no caller passes a large index.
- **Single-channel assumption:** one funding outpoint is cached (overwritten each load). Fine for the current one-connection model; note if multi-channel ever appears.
- Narrow edge (shared with the feature itself): if the app never runs during the force-close pending window, neither the sweep-txid label nor the outpoint cleanup fires until a new channel overwrites the outpoint. Consistent with the "cached as seen" design.

### Tests
No Maestro coverage (same as #59). The coop path is the more testable one — the existing coop-close flow (`remove_wallet.yaml`) could be extended to assert the closure Description on the resulting on-chain tx. `spendingTxIDs()` and the coop-detection predicate are also good unit-test targets.

### Verdict
Well-structured feature closing a real UX gap; detection design is sound and the two money-path/perf hardening items are now in. Before merge: **rebase onto #59** and **verify the txid formats match**. Base branch is `android`, consistent with the series.

PR #61, opened 2026-08-01
Onchain address management refactor

tommelters:
**Problem**
The app caches the last ten unused onchain addresses for the user to see. After each app launch, the app checks whether any of these ten have been used, and - if so - reveals new addresses. This code needed improvements.

**Updates**

- Code improvements for better efficiency and readability.
- Any potential mismatch between BDK and LDKNode, would keep the function running infinitely. This has now been capped.
- Add guards as not to crash during wallet resetting.
- In case the API returns no data or unexpected data about an address, abort the function instead of marking the address as unused.
- Improved alignment of BDK and LDKNode to make sure both have revealed the same addresses.
- Combine the 10 API calls to increase the handling speed tenfold.
- Guard against running onchain address management multiple times simultaneously.
- Fix issue where address index 0 was automatically marked as used.
- Improvements to showing the user the next available address.
- Fix issue where every time the users opens the app the available address was reset to number 1 of 10, instead showing the last address the user has chosen to see.

### DIFF
--- a/ios/bittr/AlertManager.swift
+++ b/ios/bittr/AlertManager.swift
@@ -196,6 +196,10 @@ extension UIViewController {
             yellowCard.addConstraints([messageLabelLeft, messageLabelRight, messageLabelTop])
             messageLabel.addConstraints([messageLabelHeight])
             
+            // Check whether to stack buttons vertically or horizontally.
+            let titlesFitSideBySide = buttons.allSatisfy { eachButton in eachButton.count < 16 }
+            let stackButtonsVertically = buttons.count > 2 || (buttons.count == 2 && !titlesFitSideBySide)
+            
             // Buttons stack
             let buttonsStack = UIView()
             buttonsStack.translatesAutoresizingMaskIntoConstraints = false
@@ -203,10 +207,10 @@ extension UIViewController {
             buttonsStack.backgroundColor = .clear
             yellowCard.addSubview(buttonsStack)
             let buttonsStackHeight:NSLayoutConstraint = {
-                if buttons.count < 3 {
-                    return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50)
-                } else {
+                if stackButtonsVertically {
                     return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50*CGFloat(buttons.count))
+                } else {
+                    return NSLayoutConstraint(item: buttonsStack, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: 50)
                 }
             }()
             let buttonsStackLeft = NSLayoutConstraint(item: buttonsStack, attribute: .leading, relatedBy: .equal, toItem: yellowCard, attribute: .leading, multiplier: 1, constant: 15)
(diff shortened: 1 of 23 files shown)


RubenWaterman (commented):
## Review

Note on authorship: I rebased this branch onto the current `android` during review (merge commit `1c6302b`) — #60 had squash-merged, so the PR was conflicting and predated the `.coopClose` / stale-outpoint hardening. Conflicts were resolved keeping #61's nil-safe BDK refactor (`revealAddresses`/`getAddress -> String?`) and taking android's hardened balance code; the auto-merged files were verified identical to #61's pre-merge superset (no stale identifiers, single `+4` reveal). The diff is now the 8 files of the address refactor alone.

Rewrites onchain address-pool management: batched/concurrent Esplora usage checks, a bounded BDK↔LDK derivation search, LDK/BDK reveal alignment, a re-entrancy guard, nil-safe derivation, and discard-on-corruption of the cached pool.

### Strengths
- **Bounded searches.** `revealOnchainAddresses` caps the BDK-vs-LDK match at `maxOnchainAddressSearchIndex (2500)` and Sentry-reports divergence; `alignLDKNodeRevealedAddresses` bails the moment LDK stops advancing (`nextIndex > ldkIndex`). Fixes the infinite-loop-on-mismatch.
- **Abort instead of mis-marking.** A nil/unexpected Esplora result now stops the check and keeps the cached pool rather than treating the address as unused — the most important correctness fix, since the old behavior could hand out a used address.
- **Nil-safe derivation.** `getAddress(atIndex:) -> String?` plus guards in `BoltzRefund`/`SwapManager` replace force-unwraps that could crash during a reset; callers abort cleanly.
- **Re-entrancy guard** (`isManagingOnchainAddresses`) with every exit routed through `finishOnchainAddressManagement` — I traced all paths; the flag is always reset, no stuck-flag path.
- **Pool hygiene:** `toAddresses()` discards the whole pool on parse failure or non-contiguous indices, and the persisted dict key stays `"hasBeenUsedByBittr"` so existing caches still load despite the `hasBeenUsed` rename. Batched `withTaskGroup` checks are a real speedup. Removed a stray debug `print`.

### Risks / considerations
- **BDK `Wallet` thread-safety (medium)** — `revealAddresses`/`persist`, `alignLDKNodeRevealedAddresses`, and the reveal loop run on a background queue and call BDK while light sync can touch the wallet concurrently. BDK `Wallet` isn't safe for concurrent mutation+read. Confirm all BDK access is serialized on one queue, or accept the race. The item I'd most want resolved before merge.
- **First-run pool size (low-medium).** `revealOnchainAddresses` builds the pool as every address from 0 to LDK's current index; for a restored/heavy wallet that reveals and light-sync-queries a large range, and the top-down batched check walks the full pool when used addresses sit low. Bounded by 2500; typical users unaffected.
- **`isManagingOnchainAddresses` is a plain `Bool`** touched from `manageOnchainAddresses` and `finishOnchainAddressManagement` (main.async). Fine if `manageOnchainAddresses` is only ever called on the main thread — worth confirming.
- **`alignLDKNodeRevealedAddresses`** peeks `0...(targetIndex+20)` each run (unexplained `+20` buffer) and advances LDK by ≥1 even when already aligned. Minor.

### Tests
No Maestro coverage of the address flow, and the PR's own crash fixes live in the wallet-reset / removal paths — exactly the suite flows (`remove_wallet.yaml`, `wrong_pin_*`) most likely to regress. Run those on device before merge, plus a fresh-install receive-address sanity check. `getNextUnusedAddress` / `toAddresses` (contiguity, discard) are unit-testable and worth locking down. Security: all derived addresses are the user's own wallet addresses, so no fund-misdirection surface; the Esplora check only reads `chain_stats.tx_count`.

### Verdict
Substantial, well-guarded refactor fixing real bugs (infinite loops, mis-marking used addresses, reset crashes, the "reset to address 1 of 10" issue). Now cleanly rebased on `android` with the balance hardening carried in. Before merge I'd prioritize the **BDK thread-safety** question and a **device run of the reset/removal flows**.

PR #62, opened 2026-08-03
Sentry refactoring

tommelters:
**Reporting less**
Sentry's automatic integrations were all on their defaults, so most of what arrived had nothing to act on. 
Failed HTTP requests (any 500–599), inferred watchdog terminations, and app hangs on a 2s threshold that our own node start and wallet sync exceed. 
Failures that are the user's circumstances rather than our defect: no connection, an invoice already paid, a channel that's full. Those are counted as metrics instead.

**Improvements to redacting sensitive data**
Redaction now matches addresses, invoices, LNURLs, extended public keys, txids, preimages, node pubkeys, IBANs and email addresses — plus a recovery-phrase detector.

**Metrics**
More elaborate metrics for lightning payments. And moved some metrics to the correct places. Metrics now only count in production, so a Maestro run no longer moves the numbers.

**SentryManager**
Sentry setup moved out of `didFinishLaunchingWithOptions` (150 lines of it) into `SentryManager.start()`. This new file now holds all Sentry functions.
New functions capture() and countMetric() inside SentryManager now replace most of the Sentry code that lived across other files.

**InvalidInvoice**
We've fixed a Sentry error where somehow a zero-amount invoice could be handled like a regular invoice, leading to an error.


RubenWaterman (commented):
## Review

Note on authorship: I rebased this branch onto the current `android` during review (merge `62874cf`). The merge was conflict-free per git but left one compile break — #61's address refactor added two `SentrySDK.capture(...)` calls to `ReceiveOnchain.swift`, from which this PR had removed `import Sentry`. I re-added the import as the minimal merge resolution (redaction still applies to those calls via the global `beforeSend` hook). **Follow-up:** migrate those two calls (`ReceiveOnchain.swift:127,180`) to `SentryManager.capture` for consistency with this PR's goal.

Consolidates all Sentry usage into `SentryManager` (setup, `capture`, `countMetric`) + `RedactionManager`, moves setup out of `AppDelegate`, quiets noisy automatic integrations, gates metrics to production, expands PII redaction, and fixes a zero-amount-invoice mishandling. Mostly a mechanical call-site migration (~45 files) plus three substantive pieces.

### Substantive correctness
- **Zero-amount invoice fix is a real improvement.** Replacing the stateful `temporaryIsZeroAmountInvoice` flag with the invoice as source of truth (`invoice.amountMilliSatoshis() == nil` at send time, parsed once and reused) removes the stale-flag bug class.
- **`capture(error:)` filtering** of `isConnectivityError` / `isExpectedPaymentFailure` is well-scoped; `BittrServiceError` recurses into its underlying error. Note the guard only early-returns — it does not itself emit a metric, so "counted as metrics instead" relies on each call site adding `countMetric`; worth confirming none were dropped.
- **`capture`/`countMetric` dispatch to `main.async`** — unnecessary for thread-safety and slightly risky: an error raised while main is wedged may never dispatch. Consider capturing on the calling thread.

### Redaction (security-critical)
Strong approach: `beforeSend`/`beforeSendSpan`/`beforeBreadcrumb`, plus the BIP39 phrase detector (≥12 consecutive wordlist words, separator-agnostic, checksum-independent) for the one value that must never leak. Pattern set (invoice/LNURL/xpub/bech32/base58/64–66-hex/IBAN/email/amounts) is thorough.

Gaps worth closing (defense-in-depth):
- **Non-`String` `extra` values bypass redaction** — `redacted(event:)` only handles `value as? String`; an array/dict/number `extra` carrying a sensitive string goes through untouched. Consider recursing into collections.
- **`beforeSendBreadcrumb` does not run `redactSensitiveValues()`** on the message (only tag/URL). It works today only because `Log.swift` pre-redacts its own breadcrumbs; any breadcrumb added elsewhere gets partial redaction. Adding it there would make breadcrumbs symmetric with events.
- **WIF private keys** (`5`/`K`/`L`/`c` prefix) aren't covered by the base58 pattern (anchored on `[13mn2]`); raw 64-hex keys are caught, WIF isn't.
- Confirm `stackTraceMode` isn't set to capture local variables (stack-frame locals aren't redacted; off by default).
- `device_token` is re-added to `extra` after redaction — intentional for support, but it's device-identifying; flagging as an explicit PII exception.

### Migration / tests
- ~45-file mechanical swap: low individual risk but easy to drop a context or flip error↔message. Needs a **clean compile** and a grep confirming no stray direct `SentrySDK.` outside `SentryManager` (after the merge, the only one is the ReceiveOnchain follow-up above).
- **No redaction tests** — this is exactly the pure, high-stakes logic that warrants them. A unit test feeding a known seed phrase / invoice / xpub / address / email through `redactSensitiveValues()` and asserting `[redacted]` would lock the guarantee. Strongly recommend given the downside.

### Verdict
Well-motivated cleanup with a real bug fix and a much stronger redaction story. Not a blocker, but before merge I'd add **redaction unit tests**, close the **non-string-extra / breadcrumb** gaps, migrate the two **ReceiveOnchain** calls, and **compile**. Base is `android`.

PR #63, opened 2026-08-03
Start BDK once, and keep its scan state honest

tommelters:
**Problem**
This refactor is the fix to a Sentry error BitcoinDevKit.CreateWithPersistError: Persist(errorMessage: "disk I/O error") which was the result of multiple places in the code (e.g. the CoreVC and SendVC) being able to simultaneously try to start BDK, leading to corrupt data.

**Solution**
BDK wallet starting as well as full scanning are now locked so that neither can accidentally be initiated more than once simultaneously.
Another issue was that the scan state flags weren't being cleared upon wallet reset, which has now been resolved.

### DIFF
--- a/ios/bittr/BDKManager.swift
+++ b/ios/bittr/BDKManager.swift
@@ -18,6 +18,11 @@ struct OnchainDrainPreview {
 extension BitcoinManager {
     
     func didStartBDK() -> Bool {
+        
+        // One start at a time.
+        self.bdkStartLock.lock()
+        defer { self.bdkStartLock.unlock() }
+        
         guard self.bdkWallet == nil else {
             return true
         }
@@ -66,6 +71,7 @@ extension BitcoinManager {
             self.connection = try Connection.createConnection()
         } catch {
             self.handleError(error: error, row: 211)
+            self.clearBDKState()
             return false
         }
         
@@ -73,6 +79,7 @@ extension BitcoinManager {
             self.bdkWallet = try Wallet(descriptor: bip84ExternalDescriptor, changeDescriptor: bip84InternalDescriptor, network: EnvironmentConfig.bitcoinDevKitNetwork, connection: self.connection!)
         } catch {
             self.handleError(error: error, row: 218)
+            self.clearBDKState()
             return false
         }
         
@@ -81,12 +88,20 @@ extension BitcoinManager {
             self.electrumClient = try ElectrumClient(url: EnvironmentConfig.electrumURL)
         } catch {
             self.handleError(error: error, row: 228)
+            self.clearBDKState()
             return false
         }
         
         Log.info("Did initiate wallet and blockchain.")
         SentrySDK.metrics.count(key: "sync.bdk.success")
         return true
+    }
+    
+    private func clearBDKState() {
+        // Leave nothing half-built behind after a failed start.
+        self.bdkWallet = nil
+        self.electrumClient = nil
+        self.connection = nil
     }
     
     func didSyncBdkWallet(completion originalCompletion: @escaping (Bool) -> Void) {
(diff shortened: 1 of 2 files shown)


RubenWaterman (commented):
## Review

Note on authorship: I rebased this onto the current `android` during review (merge `d49a0e4`). One conflict in `BDKManager.swift` — this PR replaces the two scan-flag assignments in the full-scan success path with `endBdkScan(scanned: true)`, while #60 (now on android) had added `storeChannelClosureTxIDIfFound()` at the same spot. Resolved to keep both (`endBdkScan(true)` is this PR's encapsulation of those two flag sets; the closure-detection call is independent). Verified repo-wide that the now-computed `bdkWalletIsScanning`/`bdkWalletHasBeenScanned`/`bdkFullScanTimedOut` have no remaining assignments — every external reference is a read.

Fixes `CreateWithPersistError: Persist("disk I/O error")` from multiple call sites starting BDK / full-scanning concurrently and corrupting the SQLite wallet. Introduces a lock-guarded `BdkScanState`, serializes `didStartBDK` behind `bdkStartLock`, makes full scans single-flight (later callers wait on the in-flight scan), clears wallet references on a failed start, and clears scan state on wallet reset.

### Strengths
- **`BdkScanState` is a clean lock-based state machine** — all mutable state behind one `NSLock`, waiters copied-then-cleared under the lock and dispatched outside it. Correct threading discipline.
- **Single-flight is right for the common path:** the claimer runs the scan; non-claimers queue as waiters and are notified via `end(scanned:)`; the claimer is never in `waiters`, so no double-notification.
- **Late-watchdog guard is thoughtful** (`markTimedOut` early-returns unless `_isScanning`; per-invocation `didComplete` makes a stale watchdog a no-op for that caller).
- **`clearBdkWalletReferences()` on each failure path** stops a half-built wallet from lingering; **minimal call-site churn** via computed proxies keeps existing readers untouched.

### Risks / correctness
- **Waiters have no timeout — the main gap.** Only the claiming caller arms the 180s watchdog; a caller that arrives mid-scan returns right after `claimBdkScan` without one. If the in-flight scan exceeds its watchdog and stalls, `markTimedOut()` sets `_timedOut` but does not drain waiters, so their completion never fires and their UI hangs until the real scan eventually calls `endBdkScan` (or forever). Suggest: on watchdog fire, notify waiters `(false)` while keeping `_isScanning` true — unblocks them without letting a new concurrent scan start (the corruption being prevented).
- **Cross-scan watchdog bleed.** Watchdogs aren't tied to a scan generation, so scan A's watchdog can fire while scan B is running and spuriously flag B as timed out (`guard _isScanning` is true for B). Low probability; a generation token checked in `markTimedOut` closes it.
- **`didStartBDK` holds `bdkStartLock` across slow I/O** (connection/wallet/electrum creation) — intended serialization, but a concurrent caller blocks fully, and on the main thread the wait is on main. Pre-existing (already synchronous); confirm callers are off-main.

### Style / tests
- The removed comment explaining why scanning stays set after a timeout ("only a restart clears it") carried real intent — worth a one-liner on `markTimedOut`. The `// TODO: higher timeout for mainnet` is a legitimate follow-up.
- No tests, but **`BdkScanState` is pure and trivially unit-testable** (claim-once, second-claimer-queues, `end` notifies all waiters with the outcome, `clear` releases with false, `markTimedOut` only while scanning). High-value given it guards against DB corruption.

### Verdict
Correct, well-structured fix for the reported corruption with good threading hygiene. The waiter-timeout gap is the one I'd address before merge (a stalled scan can hang later callers with no escape hatch); the watchdog bleed and mainnet timeout are smaller follow-ups. Now rebased on `android`.

PR #64, opened 2026-08-04
Retry LDK Node start on chain source failures

tommelters:
This is an effort to reduce the amount of FeerateEstimationUpdateFailed errors, which are usually due to the public Blockstream Electrum server rate-limiting an API call.

Now, in Node.start(), for chain source-related errors, we immediately retry starting ldkNode up to 2 times within a maximum of 10 seconds, to try and get a successful response.

### DIFF
--- a/ios/bittr/BitcoinManager.swift
+++ b/ios/bittr/BitcoinManager.swift
@@ -206,19 +206,63 @@ class BitcoinManager {
         }
         
         // Start new node.
-        do {
-            try newLdkNode.start()
-        } catch {
-            Log.info("Could not start newLdkNode. \(error)")
-            SentrySDK.capture(error: error) { scope in
-                scope.setExtra(value: "BitcoinManager row 147", key: "context")
-            }
+        guard self.didStartNode(newLdkNode) else {
             return false
         }
         
         // Successful start.
         self.ldkNode = newLdkNode
         return true
+    }
+    
+    // Start an already-built node, retrying for chain source-related failures.
+    private func didStartNode(_ node: Node) -> Bool {
+        
+        // Date and time of 1st attempt.
+        let startedAt = Date()
+        var attempt = 1
+        
+        // Start 2nd and 3rd attempts after 1 and 3 seconds, respectively.
+        let nodeStartRetryDelays: [TimeInterval] = [1, 3]
+        
+        while true {
+            do {
+                try node.start()
+                Log.info("Did start newLdkNode on attempt \(attempt).")
+                return true
+            } catch {
+                Log.info("Could not start newLdkNode (attempt \(attempt)). \(error)")
+                
+                // Check whether we've already done 3 attempts.
+                let hasAttemptsLeft = attempt <= nodeStartRetryDelays.count
+                // Only retry if less than 10 seconds have passed in total.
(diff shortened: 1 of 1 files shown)


RubenWaterman (commented):
## Review

Note on authorship: I rebased this onto the current `android` during review (merge `03925d7`). One conflict in `BitcoinManager.swift` where this PR replaces the node-start block that #62 had migrated to `SentryManager`. Resolved to keep the retry (`didStartNode`), and migrated `didStartNode`'s own `SentrySDK.capture(...)` to `SentryManager.capture(error, context: "BitcoinManager row 147")` — this was **required to compile**, since #62 removed `import Sentry` from `BitcoinManager`, and it preserves #62's migration of that exact capture. (Redaction/filtering now apply to it via `SentryManager`.)

Extracts node startup into `didStartNode(_:)`, retrying `node.start()` up to 3 attempts (initial + 2) with `[1, 3]`s backoff, gated to chain-source errors (`FeerateEstimationUpdateFailed`, `FeerateEstimationUpdateTimeout`, `ConnectionFailed`) and a 10s overall deadline — targeting Blockstream-Electrum rate-limit noise.

### Correctness (verified against ldk-node source)
- **The retry is genuinely safe.** In `Node::start()` (lib.rs), `chain_source.start()` (raises `ConnectionFailed`) and `update_fee_rate_estimates()` (raises `FeerateEstimationUpdateFailed`) both run *before* `is_running` is set true (line 642). So a thrown `start()` leaves `is_running == false` and a retry re-enters cleanly — no `AlreadyRunning`. Without this, the retry would have been a no-op.
- **No out-of-bounds:** `nodeStartRetryDelays[attempt - 1]` is only reached under `hasAttemptsLeft` (`attempt <= count`), so the index is always 0 or 1.
- **Timing is correct:** 3 attempts, `[1,3]`s sleeps, and `-startedAt.timeIntervalSinceNow < 10` caps total retrying even if individual `start()` calls are slow.
- **Noise reduction works:** Sentry capture fires only on give-up (exhausted or non-retryable); a transient failure that succeeds on retry is no longer reported.

### Risks / notes
- **`Thread.sleep` blocks the caller up to ~4s** and holds the `claimNodeStart` single-flight claim longer (waiters wait up to ~10s). `node.start()` is already synchronous/slow so this should be off-main — worth confirming.
- **Chain-source re-invocation:** after a fee-estimate failure, a retry re-runs `chain_source.start()` (which succeeded the first time). Relies on it being safe to re-call — implied by ldk-node's recoverable-`Error` contract, but the one thing I couldn't fully prove from source. Low risk.
- `isRetryableNodeStartError` overlaps `Error.isConnectivityError` (from #62) but the sets differ (`FeerateEstimationUpdateFailed` is retryable here, not connectivity there), so keeping them separate is fine.

### Tests
None. The decision logic (`isRetryableNodeStartError`, deadline math) is pure and unit-testable, but the real behavior needs a rate-limit simulation that's impractical to automate — acceptable to skip.

### Verdict
Correct, well-targeted fix; I confirmed against the dependency that the retry is actually safe against the node's run-state (the thing that could have made it a no-op). Now rebased on `android` with the capture routed through `SentryManager`.

PR #65, opened 2026-08-04
Properly report HTTP status code in alerts and to Sentry

tommelters:
**Problem**
Whenever any API called by CallsManager returned a generic HTTP error status code, this code wasn't read. So the error sent to Sentry didn't include helpful descriptions.
Additionally, the custom objects APIError and BoltzAPIError didn't conform to the localizedDescription parameter, which meant any error shown to the app user in an alert was ugly.

**Solution**
CallsManager now reads and reports the HTTP status code when it can't parse JSON from the API's reply.
APIError and BoltzAPIError now conform to localizedDescription, to show the user user-friendly error information.

### DIFF
--- a/ios/bittr/CallsManager.swift
+++ b/ios/bittr/CallsManager.swift
@@ -63,6 +63,11 @@ class CallsManager: NSObject {
                             return
                         }
                     } catch {
+                        if let statusCode = (response as? HTTPURLResponse)?.statusCode, !(200..<300).contains(statusCode) {
+                            Log.info("Request failed with status \(statusCode).")
+                            completion(.failure(.requestFailed("HTTP \(statusCode)")))
+                            return
+                        }
                         DispatchQueue.main.async {
                             SentrySDK.capture(error: error) { scope in
                                 scope.setExtra(value: "CallsManager row 60", key: "context")
--- a/ios/bittr/CallsManager.swift
+++ b/ios/bittr/CallsManager.swift
@@ -15,6 +15,19 @@ enum APIError: Error {
     case decodingFailed
 }
 
+extension APIError: LocalizedError {
+    var errorDescription: String? {
+        switch self {
+        case .invalidURL:
+            return "Invalid URL."
+        case .requestFailed(let message):
+            return message
+        case .decodingFailed:
+            return "Could not read the server's response."
+        }
+    }
+}
+
 class CallsManager: NSObject {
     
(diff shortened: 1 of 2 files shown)


RubenWaterman (commented):
## Review

Note on authorship: I rebased this onto the current `android` during review (merge `7a83daa`). One conflict in `CallsManager.swift`'s decode-`catch`, where #62 had migrated the `SentrySDK.capture` to `SentryManager.capture`. Resolved to keep this PR's HTTP-status early-return **and** #62's `SentryManager.capture(error, context: "CallsManager row 60")` for the fall-through — which also drops the old `received_data` extra (a `Data`, so it bypassed #62's string-only redaction; dropping it is the safer choice, so the concern below is now moot on this path).

Two improvements: `APIError`/`BoltzAPIError` conform to `LocalizedError` (friendly `localizedDescription` in alerts), and `makeApiCall` reports `.requestFailed("HTTP <code>")` when JSON parsing fails on a non-2xx response.

### Correctness
- **`response` scope and the check are right** — it's the dataTask closure's `response`; `(response as? HTTPURLResponse)?.statusCode` with `!(200..<300).contains` at the top of the decode-`catch` correctly separates "server returned an error body" from "2xx with malformed JSON."
- **`LocalizedError` conformance is the right mechanism** (`errorDescription` drives `localizedDescription`); `.requestFailed(message)` surfaces the message.
- 2xx-with-bad-JSON still falls through to the Sentry capture, so genuine decode bugs stay reported.

### Behavior change worth weighing
- **Non-2xx now skips Sentry — including 4xx.** For 5xx that matches #62's "server-side, not our defect" stance, but **4xx often means an app-side bug** (malformed request) and now loses Sentry visibility. Consider still reporting 4xx while suppressing 5xx.

### Style / conventions
- **Hardcoded English strings** ("Invalid URL.", "Could not read the server's response.") bypass `Language.getWord`, so they won't localize; `"HTTP 500"` shown to a user is technical. Minor consistency nit.
- The `LocalizedError` extension is duplicated verbatim across the two error types — acceptable.
- New `completion(.failure(...))` runs on the URLSession background queue like the method's other completion paths — consistent.

### Tests
None; the logic is simple and low-risk. A tiny test asserting each case's `localizedDescription` would lock the copy — optional.

### Verdict
Solid, low-risk usability/observability improvement, correctly implemented. The one judgment call is **suppressing 4xx from Sentry**; the i18n of the new strings is a minor nit. Now rebased on `android` with the capture on `SentryManager`.

PR #66, opened 2026-08-04
Give every API call a deadline

tommelters:
makeApiCall built its requests with `timeoutInterval: Double.infinity`, so a server that accepted the connection and then went quiet left the call hanging forever. Whatever was waiting on it waited forever too — a swap's Next button spinning until the user gave up, with no error to report.

Two deadlines, because one isn't enough: `timeoutInterval` is an inactivity timer that resets whenever bytes arrive, so a server dribbling out data can hold it open indefinitely. The total ceiling is `timeoutIntervalForResource`, which lives on the session configuration — hence our own URLSession, as URLSession.shared's configuration is fixed.

Callers that talk to something legitimately slow can raise the per-request timeout; values above the 60s resource ceiling need that raised too.

### DIFF
--- a/ios/bittr/CallsManager.swift
+++ b/ios/bittr/CallsManager.swift
@@ -17,9 +17,21 @@ enum APIError: Error {
 
 class CallsManager: NSObject {
     
-    static func makeApiCall(url:String, parameters:[String:Any]?, getOrPost:CallType, completion: @escaping (Result<NSDictionary, APIError>) -> Void) async {
+    // How long a request may go without producing data.
+    static let defaultTimeout:TimeInterval = 30
+    // Ceiling on a whole request, however steadily data trickles in.
+    private static let resourceTimeout:TimeInterval = 60
+    // Our own session.
+    private static let session:URLSession = {
+        let configuration = URLSessionConfiguration.default
+        configuration.timeoutIntervalForRequest = defaultTimeout
+        configuration.timeoutIntervalForResource = resourceTimeout
+        return URLSession(configuration: configuration)
+    }()
+    
+    static func makeApiCall(url:String, parameters:[String:Any]?, getOrPost:CallType, timeout:TimeInterval = defaultTimeout, completion: @escaping (Result<NSDictionary, APIError>) -> Void) async {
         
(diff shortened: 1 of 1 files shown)


RubenWaterman (commented):
## Review

Note on authorship: I rebased this onto the current `android` during review (merge `11f6d09`). It merged with no conflicts — this PR's timeout changes sit in different regions than #62's `SentryManager` migration and #65's HTTP-status check, so `CallsManager` now cleanly combines all three (session/timeout + `SentryManager.capture` + `.requestFailed("HTTP …")` + `LocalizedError`). Verified: no stray `SentrySDK`, no import breakage.

Replaces `URLRequest(timeoutInterval: .infinity)` + `URLSession.shared` with a two-layer deadline — a per-request inactivity timeout (`timeout`, default 30s) and a 60s total resource ceiling on a dedicated `URLSession` — fixing calls that hung forever when a server accepted the connection then went quiet.

### Correctness
- **Right diagnosis, right mechanism.** `timeoutIntervalForRequest` resets on each byte, so only the `timeoutIntervalForResource` ceiling truly bounds the total. Using a private static session is also necessary and correctly justified — `URLSession.shared`'s configuration is immutable, so a resource timeout can't be set on it. App-lifetime singleton, no delegate, no invalidation needed: the right pattern.
- **No regression in practice.** Every current caller is a fast JSON REST call well under 60s (Boltz status is polled over WebSocket, not through here), so the ceiling won't cut off any legitimate call. 30s inactivity / 60s total is sensible.

### Notes (non-blocking)
- The `timeout` parameter is silently capped by the fixed 60s `resourceTimeout`, so a future caller passing `timeout > 60` would be quietly ignored. Given the team's stance that no API call should exceed 60s, this is fine as a hard cap — a one-line doc-comment (or clamping `timeout` to `resourceTimeout`) would just make that explicit for the next reader.
- Setting both `request.timeoutInterval` and the session's `timeoutIntervalForRequest` is slightly redundant (the request value governs per-request inactivity); harmless.
- `async` on `makeApiCall` stays vestigial (drives a completion-handler `dataTask`) — pre-existing.

### Tests & security
- No tests; timeout behavior is impractical to unit-test reliably — acceptable.
- Security: net positive; bounding request lifetime removes an indefinite-resource-hold failure mode, no new surface.

### Verdict
Correct, valuable reliability fix with sound rationale, and safe for all current callers. Now rebased on `android`, cleanly integrated with the Sentry refactor and the CallsManager error changes.

PR #67, opened 2026-08-07
Keep LDK channel state out of device backups

tommelters:
**The problem**

- Unlike the pin and mnemonic (which are backed up to ThisDeviceOnly), all **LDK Node data** (i.e. channel state, channel manager) is backed up through **LightningStorage** and **included in syncs** across the user's devices.
- However, it's **not possible to reconstruct the users channel** on a new device. Doing so would lead to a **force-close** of the channel where the user could **lose their lightning funds**.

**The solution**

- LDK Node data should be **excluded from device backups**.
- Additionally, as a **migration for current users**, foreign LDK Node data should be recognized as such and not be reinstated when the user recovers a backup on a new device.
- Foreign LDK Node data remains valuable as a last resort to try and recover lost funds, so it gets **"quarantined"** but not removed.
- If foreign LDK Node data gets recognized, the user is **informed during wallet launch** that it's not possible to restore a Lightning connection from another device.

### DIFF
--- a/ios/bittr/AppDelegate.swift
+++ b/ios/bittr/AppDelegate.swift
@@ -144,6 +144,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         // getters also self-heal on first read, so this is just eager cleanup.
         CacheManager.migrateSecretsToKeychainIfNeeded()
 
+        // Keep Documents out of iCloud/Finder backups.
+        LightningStorage.excludeFromBackup()
+        
         UNUserNotificationCenter.current().delegate = self
         
         return true
--- a/ios/bittr/BitcoinManager.swift
+++ b/ios/bittr/BitcoinManager.swift
@@ -37,6 +37,9 @@ class BitcoinManager {
     var xpub = ""
     var coreVC:CoreViewController?
 
+    // Prevent reconstructing foreign LDK Node data
+    var didQuarantineForeignState = false
+    
     // Bittr wallet
     var bittrWallet = BittrWallet()
     
@@ -118,6 +121,9 @@ class BitcoinManager {
         
         // Delete previous LDK Node log.
         try? FileManager.deleteLDKNodeLogLatestFile()
+        
+        // Keep the channel state out of iCloud/Finder backups.
+        LightningStorage.excludeFromBackup()
         
         // Congifure LDK Node settings.
         let correctListeningAddresses = EnvironmentConfig.isDevelopment ? ["0.0.0.0:19735"] : ["0.0.0.0:9735"]
--- a/ios/bittr/Cache/CacheManager.swift
+++ b/ios/bittr/Cache/CacheManager.swift
@@ -737,7 +737,33 @@ class CacheManager: NSObject {
     // MARK: - Mnemonic
     
     static func storeMnemonic(_ mnemonic:String) throws {
+        
+        // Make sure no foreign LDK Node data remains available when creating or restoring a wallet.
+        // Foreign LDK Node data can never restore a channel, and may lead to loss of funds.
+        try quarantineLightningStateBeforeSeedImport()
+        
+        // Store new or restored mnemonic.
         try persistSecret(mnemonic, account: EnvironmentConfig.cacheKey(for: "mnemonic"), label: "mnemonic", accessibility: .afterFirstUnlockThisDeviceOnly)
+    }
+    
+    private static func quarantineLightningStateBeforeSeedImport() throws {
+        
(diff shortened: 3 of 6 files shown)


RubenWaterman (commented):
## Review

Note on authorship: during review I added one commit (`b91e1c1`, exclude the quarantine directory — see Issue 2) and rebased the branch onto the current `android` (merge `84c7057`, clean; verified the whole #59–#66 series merged with no markers and no stray `SentrySDK`).

Funds-safety fix: stop LDK channel state from being restored onto a new device (which would force-close the channel and risk fund loss). Two mechanisms — `excludeFromBackup()` (mark Documents + LDK files `isExcludedFromBackup`) and quarantine (on seed import with no prior mnemonic but existing LDK state, move it to `foreign_ldk_state/` and inform the user).

### Verified correct
- **Filename matches ldk-node.** `ldkStateFilePrefix = "ldk_node_data.sqlite"` is exactly `SQLITE_DB_FILE_NAME` (io/sqlite_store/mod.rs:28), and the builder uses it (not the `ldk_data.sqlite` default). Prefix match correctly also catches `-wal`/`-shm`.
- **Quarantine trigger is correctly scoped.** It only fires when the Keychain has no mnemonic — so it can never touch *this* device's real channel (a user with a channel has a mnemonic → early return). Move-not-delete preserves last-resort recovery.
- **Restore-seed ordering is safe.** `didStartLDK` bails without a mnemonic, and `storeMnemonic` quarantines before persisting, so the node never starts with foreign state in the normal restore-seed flow.

### Issue 1 — the legacy-migration "gap" has no safe fix; the current design is correct
I dug into the concern that `migrateSecretsToKeychainIfNeeded` → `migrateLegacyValue` writes the mnemonic directly (bypassing `storeMnemonic`'s quarantine). It *is* reachable — `c37fc6c` ("Store PIN and Mnemonic in Keychain instead of UserDefaults (#42)") confirms the mnemonic was in UserDefaults pre-#42, and UserDefaults is in iCloud backups. **But the intuitive fix (quarantine at launch when Keychain has no mnemonic + LDK state exists) is unsafe:** at first-launch, a *legit pre-#42 user upgrading in place* and a *foreign pre-#42 backup restored to a new device* are byte-for-byte identical (`Keychain{} + UserDefaults{mnemonic,pin} + LDK state`), because pre-#42 devices wrote nothing device-local and ThisDeviceOnly Keychain (the only non-restored store) is empty for both. So a launch quarantine would force-close the legit upgrader's own channel — worse than the bug.

Conclusion: the PR's `storeMnemonic`-only trigger is the right choice (an in-place upgrader never calls it), and the post-#42 restore case is already handled (no mnemonic anywhere → node can't start → user re-imports → quarantine fires). The only residual gap — a pre-#42-backup foreign restore — is fundamentally undistinguishable at launch and a shrinking population (`excludeFromBackup` + #42 keep it out of any refreshed backup). If you want to cover it, the only safe route is a product-level "was this device restored from a backup?" prompt at migration; not something to bake in silently. **No code change made here.**

### Issue 2 — applied
- **Fixed (`b91e1c1`):** the quarantine dir `foreign_ldk_state/` wasn't matched by the state-file prefix, so `excludeFromBackup` never covered it — quarantined foreign data could be re-synced and re-restored, defeating the quarantine. Now excluded explicitly right after creation.
- **Still to verify (not code-fixable cheaply):** `excludeFromBackup` runs before the SQLite `-wal`/`-shm` siblings exist, so effectiveness rests on directory-level exclusion of the top-level Documents dir — which is technically supported but discouraged. Worth an **on-device check** that a real backup actually skips `ldk_node_data.sqlite*`; if it doesn't, the proper fix is moving LDK state into a subdirectory and excluding that (a larger change).

### Minor
- `excludeFromBackup` excludes all of Documents (also BDK cache) — acceptable, since BDK data is re-derivable from the seed.
- The `quarantineLightningStateBeforeSeedImport` comment says an unreadable Keychain means "don't touch anything," but it actually *throws* out of `storeMnemonic` (aborts the import). Fail-closed is fine; the comment misdescribes it.
- `// MARK: -` lines aren't indented; line-23 comment has a stray-whitespace artifact. Cosmetic.

### Tests
No tests, and this is the PR that most warrants them: `hasLightningState`, the quarantine move, and the filename/prefix match are cheap to unit-test, and a manual/Maestro backup-restore-with-channel scenario asserting "quarantined, no force-close" would be high-value given the stakes.

### Verdict
Correct core mechanism, filename verified, and the common paths are safe. The Issue-1 concern turned out to be unfixable-safely (the design is right as-is), and the Issue-2 quarantine-dir hole is now closed. Before merge: **verify on-device that the Documents exclusion actually keeps `ldk_node_data.sqlite*` out of a backup** — that's the load-bearing assumption for the whole PR.

PR #68, opened 2026-08-07
Validate onchain addresses with BDK instead of a regex

tommelters:
**Improvements to SendVC onchain address validation**

- Previous regex did not recognize mainnet P2SH entries. This is now fixed.
- Testnet and Regtest addresses were being accepted on Mainnet. This is now fixed.
- Regex did not recognize case-sensitive Base58 addresses. This is now fixed.
- We now use BDK's built-in address validation rather than a custom regex.

### DIFF
--- a/ios/bittr/Move, Send, Receive/SendVC/Send/AddressParsing.swift
+++ b/ios/bittr/Move, Send, Receive/SendVC/Send/AddressParsing.swift
@@ -6,6 +6,7 @@
 //
 
 import UIKit
+import BitcoinDevKit
 import LNURLDecoder
 import LightningDevKit
 import Sentry
@@ -16,7 +17,7 @@ extension SendViewController {
         print("Code: " + code)
         
         // Parse code components.
-        let bitcoinAddress = code.lowercased().extractBitcoinAddress()
+        let bitcoinAddress = code.extractBitcoinAddress()
         let lightningInvoice = code.lowercased().extractLightningInvoice()
         let lnurl = code.lowercased().extractLNURL()
         let amount = code.lowercased().extractAmount()
@@ -87,8 +88,10 @@ extension String {
     
     func extractBitcoinAddress() -> String? {
         let components = self.components(separatedBy: CharacterSet(charactersIn: "&:?="))
-        for eachComponent in components where eachComponent.isValidBitcoinAddress() {
-            return eachComponent
+        for eachComponent in components {
+            if let address = eachComponent.asBitcoinAddress() {
+                return address
+            }
         }
         return nil
     }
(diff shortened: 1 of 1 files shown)


RubenWaterman (commented):
## Review

Note on authorship: I rebased this onto the current `android` during review (merge `c93b5e3`, clean — the #59–#67 series merged with no markers and no stray `SentrySDK`).

Replaces a hand-rolled onchain-address regex with BDK's `Address(address:network:)` validation and stops lowercasing the whole scanned code before extracting the address. Fixes mainnet P2SH not recognized, testnet/regtest accepted on mainnet, and case-sensitive Base58 mis-handling.

### Correctness
- **BDK validation is the right call and fixes all three bugs at once** — a real base58/bech32 decode + checksum + network check, so it accepts every valid type for the configured network (including taproot `bc1p`, which the old `bc1[qzp]` regex missed) and rejects wrong-network addresses. Confirmed `isValidBitcoinAddress`/`asBitcoinAddress` are used only within this file, so no other flow's behavior shifts, and `EnvironmentConfig.bitcoinDevKitNetwork` exists.
- **The case-handling holds up.** `asBitcoinAddress()` tries the string as-is first (preserving Base58's case-sensitive checksum — the whole reason the blanket `.lowercased()` was a bug), then lowercases only when there are no lowercase letters (all-uppercase Bech32). The `rangeOfCharacter(from: .lowercaseLetters) == nil` guard correctly refuses to "fix" a mixed-case string — mixed-case Bech32 is invalid per BIP173, and lowercasing a Base58 corrupts its checksum. No false accepts, no false rejects: every valid form passes at step 1 or step 3.
- Only the bitcoin-address extractor dropped `.lowercased()`; the lightning/LNURL/amount extractors keep it — correct, those are case-insensitive.

### Minor
- All-uppercase Bech32 returns the lowercased canonical form — fine, since the same BDK lib validates and later builds the tx.
- Pre-existing (not this PR): `print("Code: " + code)` in `parseCode` logs the full scanned string (address/invoice/LNURL) to the console — worth removing for privacy while nearby.
- Perf: a BDK parse per URI component instead of a regex — negligible on paste/scan.

### Tests & security
- No tests, but `asBitcoinAddress`/`isValidBitcoinAddress` are pure and cheap to unit-test — a table (mainnet P2PKH/P2SH/bech32/taproot accepted; testnet/regtest rejected on mainnet; uppercase-bech32 accepted; mixed-case and corrupted-checksum rejected) would lock these fixes down and guard a money path. High value for the cost.
- Security: net positive — rejecting malformed and wrong-network addresses reduces the chance of sending funds to an unintended destination.

### Verdict
Solid, correct, well-scoped — swapping bespoke regex for the reference library is exactly right, and the case-handling edge cases check out. Only real gap is the absent (cheap) unit tests. Now rebased on `android`.

